PR #10, opened 2014-11-26
Define our own strlcpy if one doesn't exist

Keithbsmiley:
Should fix https://github.com/thoughtbot/pick/issues/6

**Needs to be tested on bsd**

PR #16, opened 2014-12-13
Use Autoconf to check for strlcpy

ecerulm:
Autoconf can check for strlcpy. If it finds it without having to use
a library it won't add anything to LIBS. Otherwise it will try
libbsd and if finds it there it will link libbsd.

Use also AC_CHECK_HEADERS to only include bsd/string.h if it actually
exists.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,6 +9,11 @@ AC_TYPE_SSIZE_T
 AC_FUNC_MALLOC
 AC_FUNC_REALLOC
 AC_CHECK_FUNCS([memmove setlocale strdup])
+AC_SEARCH_LIBS([strlcpy], [bsd], [], [
+  AC_MSG_ERROR([unable to find strlcpy() function])
+])
+
+AC_CHECK_HEADERS([bsd/string.h])
 
 # sys/queue.h exists on most systems, but its capabilities vary a great deal.
 # Test for required macros.
--- a/configure.ac
+++ b/configure.ac
@@ -10,7 +10,7 @@ AC_FUNC_MALLOC
 AC_FUNC_REALLOC
 AC_CHECK_FUNCS([memmove setlocale strdup])
 AC_SEARCH_LIBS([strlcpy], [bsd], [], [
-  AC_MSG_ERROR([unable to find strlcpy() function])
+  AC_MSG_ERROR([unable to find strlcpy() function. Try to install libbsd])
 ])
 AC_SEARCH_LIBS([initscr], [ncurses], [], [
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -9,6 +9,14 @@ AC_TYPE_SSIZE_T
 AC_FUNC_MALLOC
 AC_FUNC_REALLOC
 AC_CHECK_FUNCS([memmove setlocale strdup])
+AC_SEARCH_LIBS([strlcpy], [bsd], [], [
+  AC_MSG_ERROR([unable to find strlcpy() function])
+])
```
calleerlandsson: If neither `strlcpy` nor libbsd is present on the system, will this error out? #10 bundles implementations of `strlcpy` and `strlcat` which are used if `AC_CHECK_FUNCS` doesn't find the functions. Would that do as a replacement for `AC_SEARCH_LIBS`?

src/ui.c
```diff
@@ -45,7 +50,7 @@ start_curses()
 	int fd;
 
 	signal(SIGINT, int_handler);
-	freopen("/dev/tty", "r", stdin);
+	assert(freopen("/dev/tty", "r", stdin));
```
calleerlandsson: Would using an `if` statement to check the return value and `err` to error if `freopen` returns `NULL` improve consistency with other parts of the code base?

calleerlandsson:
Thanks for contributing @ecerulm! I'm sorry for not getting back to you earlier. I've been travelling without internet access for a month.

I left some line comments.

configure.ac
ecerulm: Yes it will error out,  I guess the message should be "unable to find strlcpy(). Make sure libbsd is installed"  

Guess if #10 gets merged you can remove the whole `AC_SEARCH_LIBS`  since you are not dependent on `libbsd`  after #10. 

This was more  to solve the current issue of warning the user that there was no `strlcpy` at `./configure`  instead of at compile time. It depends on when are you going to merge #10, you can always remove this `AC_SEARCH_LIBS` after merging #10.

ecerulm:
I updated the error message if `strlcpy` cannot be found prompting the user to install `libbsd`. #10 will make that obsolete when it's merged but until that happens it's needed to build in Linux. 

I replaced the `assert` with `if/err` construct.

calleerlandsson:
Thanks @ecerulm. I will squash and merge this.

calleerlandsson:
Merged as 59c434f1ecc78ccfd6ba53de37f7ef9b6387624a and 0c03b0d76a8546016f824f7f977593fcbf1dab90.

Issue #20, opened 2014-12-22
No indication of cursor in pick's text entry


ggilder:
When entering text in pick, one can use the left/right arrow keys to insert and delete characters in the middle of the entry. However, there's no visual indication of the cursor, so it's very difficult to do this accurately. :wink:

calleerlandsson:
Thanks for reporting @ggilder. I'm able to reproduce this when calling `pick` using `system()` from within Vim. When executing `:!pick` in Vim and when I run `pick` from my shell I do get a cursor, though. Could you confirm that this is the case for you too?

ggilder:
Yup, I get a cursor in the shell. Weird!

calleerlandsson:
Thanks @ggilder. I think this is related to #19, #18, and #4.

PR #23, opened 2015-01-24
Show the cursor position

Keithbsmiley:
This fixes https://github.com/thoughtbot/pick/issues/20

There are probably some edge cases here so if people could test this out that would be awesome.


Keithbsmiley:
Just realized I'm not sure if this is needed anymore. Headed out will close if so.

Issue #21, opened 2014-12-22
Highlight matched portion of items


ggilder:
Other fuzzy finders like [ctrlp](https://github.com/kien/ctrlp.vim) and [fzf](https://github.com/junegunn/fzf) highlight the matched portion of items, i.e. if I have an item `README.md` in the input list and my search is `mm`, the substring `ME.m` will be highlighted in a different color. This can be useful visual feedback when searching large numbers of entries, to see why a particular search is matching something unexpected.

calleerlandsson:
I would also like this. PR's are welcome.

Keithbsmiley:
WIP on https://github.com/thoughtbot/pick/tree/ks-underline

Keithbsmiley:
Works, just needs to be cleaned up.

Keithbsmiley:
See https://github.com/thoughtbot/pick/pull/24

calleerlandsson:
This feature was added by @mptre in #97. Thanks!

Issue #4, opened 2014-11-24
Running pick using system() in Vim hinders Vim from restoring the terminal on exit.


calleerlandsson:
When I run `pick` from within Vim using the code below, when exiting Vim, the terminal state is not restored. Instead, the screen Vim drew is still visible. 

```
function! PickCommand(choices_command, pick_args, vim_command)
  try
    let selection = system(a:choices_command . " | pick " . a:pick_args)
  catch /Vim:Interrupt/
    redraw!
    return
  endtry
  redraw!
  exec a:vim_command . " " . selection
endfunction

nnoremap <leader>p :call PickCommand("find * -type f", "", ":e")<cr>
```

It would be great if someone else could replicate this. I don't know if this is a bug in pick, Vim or my terminal emulator.

... It's probably in pick.

teoljungberg:
An idea that @calleerlandsson and I talked about IRL is to explicitly reset the terminal when closing ncurses. Right now we configuring ncurses on boot: https://github.com/thoughtbot/pick/blob/0c03b0d76a8546016f824f7f977593fcbf1dab90/src/ui.c#L46-L65
Maybe doing the same on when we're stopping ncursise might do the trick?

ggilder:
This is actually still happening for me... do you think there are any changes to the vim function needed? Here's what I'm using:

```
" pick
function! EscapeFilePath(path)
  return substitute(a:path, ' ', '\\ ', 'g')
endfunction

function! Pick(choice_command, vim_command)
  try
    let selection = system(a:choice_command . " | pick")
  catch /Vim:Interrupt/
    " Swallow the ^C so that the redraw below happens; otherwise there will be
    " UI leftovers on the screen
    redraw!
    return
  endtry
  redraw!
  if strlen(selection) > 0
    exec a:vim_command . " " . EscapeFilePath(selection)
  endif
endfunction

nnoremap <leader>t :call Pick("find * -type f", ":e")<cr>
```

Keithbsmiley:
@ggilder see https://github.com/thoughtbot/pick/issues/32 you need to use `pick -X`

PR #24, opened 2015-01-24
Underline portion of matches

keith:
screen shot 2015-01-23 at 20 17 47

This implements underlining the portion of matches that make them order it. I'd love some other people to look over and try this to catch the edge cases. As you can see in the screenshot there is an off by one error somewhere so the bottom match isn't underlined correctly. We may also prefer some other color changes here but whatever everyone wants to do is fine.

### DIFF
--- a/src/choices.c
+++ b/src/choices.c
@@ -14,7 +14,7 @@
 #include "choices.h"
 
 size_t
-min_match_length(char *str, char *query, size_t *match_start_po)
+min_match_length(char *str, char *query, size_t *start_pos)
 {
 	size_t mlen;
 	size_t mstart;
@@ -22,31 +22,27 @@ min_match_length(char *str, char *query, size_t *match_start_po)
 	size_t mpos;
 	int matched = 0;
 
-	for (mlen = 0, mstart = 0; str[mstart] != '\0'; ++mstart) {
+	for (mlen = 0, mstart = 0; str[mstart] != '\0'; ++mstart)
 		if (tolower(str[mstart]) == tolower(query[0])) {
 			if (matched == 0) {
 				matched = 1;
-				*match_start_po = mstart;
+				*start_pos = mstart;
 			}
 
-			for (qpos = 1, mpos = mstart + 1; query[qpos] != '\0'; ++qpos) {
+			for (qpos = 1, mpos = mstart + 1; query[qpos] != '\0'; ++qpos)
 				for (;; ++mpos) {
-					if (str[mpos] == '\0') {
-						return mlen;
-					}
+					if (str[mpos] == '\0')
+            return mlen;
 
 					if (tolower(str[mpos]) == tolower(query[qpos])) {
 						++mpos;
 						break;
 					}
 				}
-			}
 
-			if (mlen == 0 || mlen > mpos - mstart + 1) {
-				mlen = mpos - mstart + 1;
-			}
+			if (mlen == 0 || mlen > mpos - mstart + 1)
+        mlen = mpos - mstart + 1;
 		}
-	}
 
 	return mlen;
 }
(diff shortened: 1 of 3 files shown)


Keithbsmiley:
Closes https://github.com/thoughtbot/pick/issues/21 when it's merged.

src/choices.c
```diff
@@ -23,6 +23,9 @@ min_match_length(char *str, char *query)
 
 	for (mlen = 0, mstart = 0; str[mstart] != '\0'; ++mstart)
 		if (tolower(str[mstart]) == tolower(query[0])) {
+			if (*start_pos == -1)
+				*start_pos = mstart;
```
calleerlandsson: I'm not sure I understand how this works. When is `*start_pos == -1`?

src/choices.c
```diff
@@ -14,7 +14,7 @@
 #include "choices.h"
 
 size_t
-min_match_length(char *str, char *query)
+min_match_length(char *str, char *query, int *start_pos)
```
calleerlandsson: Now that this methods also communicates back the start position of the match would renaming it `find_match`, accepting another pointer for the length and returning `1` if a match was found and `0` if not, make this clearer?

src/choices.c
```diff
@@ -39,31 +42,33 @@ min_match_length(char *str, char *query)
...
+	return (float)qlen / (float)*mlen / (float)slen;
 }
 
-
 void
 choices_score(struct choices *cs, char *query)
 {
 	struct choice *c;
+	int mpos = 0;
+	size_t mlen = 0;
```
calleerlandsson: I see how `mpos` is reset on every iteration in `score_str` but I don't get if `mlen` will reset as well? It seems like it will only reset if it is `-1`?

src/ui.c
```diff
@@ -85,12 +85,20 @@ int_handler()
 }
 
 void
-put_line(int y, char *str, int len, int so)
+put_line(int y, char *str, int len, int so, int mpos, int mlen)
```
calleerlandsson: Since this function is not only used for putting matcher would it be clearer that this was mure general purpos if `mpos` and `mlen` were renamed `upos` and `ulen` for underline?

calleerlandsson:
This is very cool @Keithbsmiley! I think @halogenandtoast ran into the same off by one error when setting up tests for `min_match_length`. `min_match_length` is very complex, do you think we could pull out some things or replace some of the looping with recursion?

teoljungberg:
This looks different when you call `pick` from inside of `vim`

screenshot 2015-01-24 16 52 53

Keithbsmiley:
Ugh on the difference through vim. I hope that's related to #4

src/choices.c
Keithbsmiley: This is just a flag to test if `start_pos` has already been set. The issue here is that it could be reset if the word being searched contained the first character of the query more than once.

src/choices.c
Keithbsmiley: I don't have a preference on the naming but the implementation would change in that case, since right now this still returns the same length but then the index of the first character of the match, not whether or not there was one.

src/ui.c
Keithbsmiley: Ah yes I could do that. Although I'm also not sure underline is the best long term case for this, it was just the easiest to test with during development.

teoljungberg:
@keithbsmiley it's also present in bash, so it's not vim only

Keithbsmiley:
Oh interesting, I'm also seeing this on a different machine still in zsh...

calleerlandsson:
Hey @keith, thanks for spending time on this! Quite a lot has changed on master since this was opened. Would you like to update this PR to get it merged in?

keith:
Closing for now, there were still some issues with compatibility with this. Feel free to try and take it over!

PR #25, opened 2015-01-24
Add project specific vimrc

Keithbsmiley:
While contributing to this project I had a lot of trouble getting vim
setup correctly. Whether or not we check this in I think we should
provide these settings to other developers so they don't have issues
with indentation and such.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -10,6 +10,16 @@ To build from a fresh checkout:
     ./configure
     make
 
+Editing
+-------
+
+Use these Vim settings for the correct indention and formatting:
+
+```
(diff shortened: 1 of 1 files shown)


calleerlandsson:
I think this is a great idea. Maybe we could add it to `DEVELOPING.md` like @mike-burns has done in [argonaut](https://github.com/bitptr/argonaut/blob/0f586c6e2c37d88bf6306aa5439b433b7d0becea/DEVELOPING.md)?

Keithbsmiley:
Moved this info to `DEVELOPING.md`

calleerlandsson:
Looks good to me!

Issue #19, opened 2014-12-22
vim-tmux-navigator stops working when pick runs inside vim


ggilder:
I can still switch tmux panes while pick runs in a shell, but if I run vim and then bring up pick inside it, I can no longer navigate to another pane using the [vim-tmux-navigator](https://github.com/christoomey/vim-tmux-navigator) key bindings. I'm guessing this might be because the value of `pane_current_command` changes while pick is up?

calleerlandsson:
I think this is an effect of #4.

calleerlandsson:
@ggilder Could you test this and see if this issue persists after merging #26?

Keithbsmiley:
It looks like this isn't a problem anymore but it would still be good if @ggilder could verify.

ggilder:
Actually, I am still seeing the problem as of 94180de913e0462ba856d592bb38a7a2f71f2952...

calleerlandsson:
@ggilder are you running with the `-X` flag?

calleerlandsson:
Since merging #35 (just now) you should not have to pass the `-X` flag anymore.

calleerlandsson:
@ggilder Does this issue persist?

ggilder:
Tested with 1.1.0, and unfortunately I still see the issue. I tried both with `-X` and without.

calleerlandsson:
@ggilder Interesting. What version of Mac OS are you running? I wonder what the difference between your system and @Keithbsmiley's is.

Am I understanding this correctly? This is not really an issue with vim-tmux-navigator (the Vim plugin) since Vim is no longer receiving input. Instead it is a problem with pick and tmux key bindings. When run from the shell in tmux pick does not affect tmux key bindings but when run from vim inside a shell in tmux, tmux key bindings do not work.

ggilder:
@calleerlandsson I'm running 10.9.5; I wonder if it might work on 10.10. I don't have a 10.10 system handy to test on unfortunately.

But yes, I think the problem is as you describe: while pick is up in Vim, Vim does not receive input so vim-tmux-navigator hotkeys don't work.

Keithbsmiley:
Ah yes I misunderstood what was happening. Pressing `CTRL-J` while pick is open definitely doesn't do what it should in tmux.

calleerlandsson:
[...]

Issue #18, opened 2014-12-22
Keycode sent by arrow keys in vim changes after using pick


ggilder:
Below is the result of entering the key codes by hitting `^V` and each arrow key (up, down, right, left) in vim.

```
fresh vim startup, before running pick:
^[0A^[0B^[0C^[0D

after running pick and exiting either successfully or via interrupt:
^[[A^[[B^[[C^[[D
```

This messes up any vim mappings set for the arrow keys.

teoljungberg:
This happens to me aswell, I compiled `pick` from the `ks-bsd` branch.

calleerlandsson:
I think this is caused by #4.

PR #26, opened 2015-02-20
Solve Vim `system()` related issues.

calleerlandsson:
- Replace ncurses with tputs and termios.
- Add an `-X` option to disable use of the alternate screen.

Hopefully solves #4, #18, #19 and #20.

@Keithbsmiley, @mike-burns, @teoljungberg  and @ggilder, could you guys help test this on your machines?

### DIFF
--- a/src/ui.c
+++ b/src/ui.c
@@ -1,5 +1,7 @@
 #include "config.h"
+#include <termios.h>
 #include <curses.h>
+#include <term.h>
 #include <err.h>
 #include <fcntl.h>
 #include <locale.h>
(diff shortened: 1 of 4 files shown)


calleerlandsson:
I have some refactoring to do but wanted to get this tested as soon as possible.

teoljungberg:
I'll give it a whirl for the next few days, I can already confirm that it doesn't fuckup the arrow-keys

src/ui.c
```diff
@@ -43,38 +45,44 @@
...
+FILE *tty;
+struct termios oldattr;
+int using_alternate_screen;
 
 void
 start_curses()
 {
-	int fd;
+	struct termios newattr;
+	if ((tty = fopen("/dev/tty", "w+")) == NULL) {
```
mike-burns: How about a newline between the declarations and the start of the code (the `if`)?

mike-burns:
Cool progress. Some general code cleanup will be nice (more vertical spacing/newlines), but it's certainly off to a fun start.

calleerlandsson:
After some testing, I realize that this breaks using arrow keys in pick. This is due to the fact that `getc` reads a character where as ncurses `getch` when using `keypad(stdscr, TRUE);` reads multiple characters in the case of function keys. I will look into how we can replicate this functionality.

calleerlandsson:
I have pushed a solution to the arrow key issue. I have no idea how portable it is.

teoljungberg:
@calleerlandsson The arrow keys work in the shell now, but not from vim

calleerlandsson:
Damn. I can reproduce it too.

calleerlandsson:
If someone wants to tackle this, I think the solution is in logging read characters when pressing arrow keys while running within Vim. If not, I'll do it on Thursday.

calleerlandsson:
It's now working and I can build and run successfully on Yosemite and Arch. I want to do a lot of refactoring in general but would prefer to do this in a separate PR.

mike-burns:
* Works on Debian Jessie, in and outside of vim, including arrow keys.
* Works on OpenBSD 5.6-RELEASE (inside a VM), in and outside of vim, including arrow keys.
* Works on FreeBSD 10.1-RELEASE (inside a VM), in and outside of vim, including arrow keys.

LGTM!

calleerlandsson:
Awesome! Thanks for testing this so thoroughly, @mike-burns! Merging.

PR #35, opened 2015-02-27
Automatically disable alternate screen in Vim.

calleerlandsson:
Detect running from within Vim and automatically disable usage of the
alternate screen.

### DIFF
--- a/man/man1/pick.1
+++ b/man/man1/pick.1
@@ -42,12 +42,7 @@ searching.
 .It Fl o
 output description of selected on exit.
 .It Fl X
-disable the use of the alternate screen terminal feature. This is required when running
-.Nm pick
-using
-.Fn system
-from within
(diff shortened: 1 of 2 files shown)


mike-burns:
I like that there is still the option to force it with `-X`. I'll submit a PR to override it with `-x`, later. LGTN.

calleerlandsson:
I agree. Thanks @mike-burns! Since it isn't that much code I'll add it in this PR.

calleerlandsson:
This is ready for another pass.

mike-burns:
Can you update the manpage to document that it also looks for `$VIM`?

calleerlandsson:
Good point. Done in b1d19e4.

mike-burns:
Looks good to merge.

calleerlandsson:
Great! Merging.

PR #36, opened 2015-02-27
Update README.md.

calleerlandsson:


mike-burns:
Will you be signing the tarball?

calleerlandsson:
I will and am signing tarballs. Should I add instructions to verify the trails signature?

mike-burns:
Yes, I think that would be a nice addition.

mike-burns:
Let's Go To Malmö.

calleerlandsson:
Going to Malmö.

PR #40, opened 2015-03-02
Use the sha of the actual release tarball.

calleerlandsson:
The sha of the tarball generated by `make distcheck` is different on every
run because of the file modification time of files in the archive.

Downloading the actual release tarball and using the sha for that in the
Homebrew formula and AUR PKGBUILD ensures that they contain the correct sha.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -55,5 +55,7 @@ release_push_arch:
 release_clean_arch:
 	rm -rf release-arch
 
-DIST_SHA: Makefile distcheck
+DIST_SHA: Makefile
+	rm $(srcdir)/$(DIST_ARCHIVES)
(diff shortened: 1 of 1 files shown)


teoljungberg:
LGTM

mike-burns:
I was thinking about this over the weekend.

Depending on `distcheck` means that it is always rebuilt. Is there another target we can depend on that invokes `distcheck` only if needed? What if you depended on `$(srcdir)/$(DIST_ARCHIVES)` or something like that? Would that also solve this issue?

calleerlandsson:
@mike-burns That would work but I think that would mean that only the person who generated the original release tarball would be able to create the formula and PKGBUILD. Also if that person deleted the tarball, they would not be able to generate the formula and PKGBUILD.

mike-burns:
OK yes, I see the issue.

Ultimately this comes down to what your workflow is. If the same person generates the tarball, formula, and PKGBUILD at the same time, then my idea makes some sense; if it's two or three different people, then this PR LGTM.

calleerlandsson:
Thanks for reviewing. I'll merge this.

PR #41, opened 2015-03-05
Link to pick.vim.

calleerlandsson:


README.md
```diff
@@ -61,17 +61,12 @@ cd $(find . -type d | pick)      # Select a directory to cd into
 Pick can also easily be used from within Vim with this VimL function:
 
 ```viml
-function! PickCommand(choice_command, pick_args, vim_command)
-  try
-    let selection = system(a:choice_command . " | ./src/pick" . a:pick_args)
-  catch /Vim:Interrupt/
```
teoljungberg: Was this only neccessary because `selecta` didn't use the alternate screen thing, and pick does?

calleerlandsson: I actually do not know. My theory is that selecta handles control characters differently.

teoljungberg: If you try it out and fire up `pick` from inside of vim and then ctrl-c you'll see your screen is borked

teoljungberg: So the function is needed as is

README.md
```diff
@@ -63,15 +63,15 @@ Pick can also easily be used from within Vim with this VimL function:
 ```viml
 function! PickCommand(choice_command, pick_args, vim_command)
   try
-    let selection = system(a:choice_command . " | ./src/pick" . a:pick_args)
+    let selection = system(a:choice_command . " | pick" . a:pick_args)
+    if v:shell_error == 0
+      exec a:vim_command . " " . selection
+    endif
   catch /Vim:Interrupt/
     " Swallow the ^C so that the redraw below happens; otherwise there will be
```
mike-burns: Is this comment still valid?

mike-burns: I guess it is.

calleerlandsson: Yes. I mistakenly removed this in an earlier commit since it had no effect on my Arch Linux machine. This made it fail on @teoljungberg's Mac. Instead of returning we fall through to the redraw on line 74.

mike-burns:
The code works from vim.

Can you add something to warn people that it only works from terminal vim, not GUI vim.

README.md
```diff
@@ -58,31 +58,18 @@ git ls-files | pick | xargs less # Select a file in the current git repository t
...
-
-Then map keys to call this function:
+Pick can also easily be used from within Vim both using `system()` and `!`. For
+ready-to-map functions, see [the pick.vim Vim plugin]. For an examples of how to
+call `pick(1)` from within Vim, see [the pick.vim source code].
 
-```viml
-nnoremap <leader>e :call PickCommand("git ls-files", "", ":e")<cr>
-```
+***Please note:*** pick requires a fully functional terminal to run and can
```
gabebw: "therefore cannot"?

calleerlandsson: Good catch. Thanks @gabebw.

calleerlandsson:
I will merge this.

PR #43, opened 2015-03-09
Open `/dev/tty/` in read and append mode. 

### LINKED ISSUE #42: Illegal seek
Giving pick a list of 200.000 strings and then typing really fast into pick's search field makes pick crash with the error message `pick: getc: Illegal seek`.

To reproduce run the following and type really fast:

```sh
cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 200000 | pick
```

calleerlandsson:
This avoids illegal seek errors when users send input while we're
outputting to the terminal.

Closes #42.

### DIFF
--- a/src/ui.c
+++ b/src/ui.c
@@ -72,7 +72,6 @@ start_ui()
 	tcgetattr(fileno(tty), &oldattr);
 	newattr = oldattr;
 	newattr.c_lflag &= ~(ICANON | ECHO);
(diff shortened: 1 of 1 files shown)


mike-burns:
Nice catch, very subtle.

calleerlandsson:
Ha. Thanks. Merging.

PR #49, opened 2015-03-18
Add ncursesw compatibility

eavgerinos:


calleerlandsson:
Hey @eavgerinos, thanks for submitting this PR!

I pulled down this branch and tried piping som unicode characters to pick, tried typing some unicode characters in the search field, tried selecting one of the unicode choices, and found some problems:

- pick will not accept unicode characters as input to the search field but will simply ignore them. This is because we're reading a normal character and are checking for a range of printable ASCII characters.
- When selecting a choice with unicode characters in it, the characters get mangled when output to stdout when pick is exiting. I think this is because we're still storing the strings as `char` arrays and output them with `printf`.

Am I misunderstanding the purpose of this PR or is it just not working on my system?

eavgerinos:
It's true, it does not fix anything about unicode characters. The nature of this PR is to give the ability to build and link to the libncursesw if it is available on your system instead of demanding libncurses.

calleerlandsson:
I knew there was something I did not quite get! Now I do, though. Sounds like a great idea!

I'll merge this.

configure.ac
```diff
@@ -46,5 +46,12 @@ AC_TRY_LINK([
 AC_DEFINE(HAVE_FULL_QUEUE_H,, Define if sys/queue.h includes required macros)
 ])
 
+AC_TRY_LINK([
+	#include <ncursesw/curses.h>
+	#include <ncursesw/term.h>
+], [], [
+AC_DEFINE(HAVE_NCURSESW_H,, Define if ncursesw is available)
+])
```
calleerlandsson: I just realized, won't this be handled by `autoconf`? After building this locally I have a modified `config.h.in` with:

```diff
+/* Define if ncursesw is available */
+#undef HAVE_NCURSESW_H
+
```

eavgerinos: Yeap, sorry, forgot to stage it.

Will do it later tonigh! Good catch! thanks

calleerlandsson: Awesome. Thanks!

eavgerinos: @calleerlandsson  Sorry for being late. Fixed :smiley:

calleerlandsson: @eavgerinos No stress! Thanks for fixing it!

I'm still curious, though, what does the `AC_TRY_LINK` construct do that the `AC_SEARCH_LIBS` and `config.h.in` does not?

eavgerinos: I'm just using it do define the constant `HAVE_NCURSESW_H` in order for the compiler to decide what it should link.

calleerlandsson: Won't that be handled by `AC_SEARCH_LIBS` on configure when `config.h` is created?

eavgerinos: This is done because we need to include different headers, so I use the `HAVE_NCURSESW_H` hdef to decide the header.

calleerlandsson: I'm sorry if I'm not getting this correctly, but won't `AC_SEARCH_LIBS` define `HAVE_NCURSESW_H` in `config.h` if `ncursesw` exists, that will then be available in `src/ui.c`?

What I'm suggesting is we drop `AC_TRY_LINK`. Does that build on your system?

eavgerinos: Yeap, it builds (it's also a [patch](https://github.com/eavgerinos/pkg-pick/blob/master/debian/patches/ncursesw-compatibility) I included in the debian package).

The [`AC_TRY_LINK`](http://ftp.gnu.org/old-gnu/Manuals/autoconf-2.53/html_node/Examining-Libraries.html) defines the `HAVE_NCURSESW_H` hdef.

On the other hand [`AC_SEARCH_LIBS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Libraries.html) iterates through a list of alternative choices in order to tell if at least 1 of the alternatives is available on the system.

So, imho, we need `AC_TRY_LINK`

calleerlandsson: Great, I'll merge this.

PR #52, opened 2015-03-24
Fix undefined behavior in src/choices.c

GBuella:
using libc shipped with cygwin,
 the build breaks with the following error message:

choices.c:25:3: error: array subscript has type 'char' [-Werror=char-subscripts]
   if (tolower(str[mstart]) == tolower(query[0])) {

Note: implementing tolower as a macro that does a lookup in
 an array - using the argument as an index - is a valid
 implementation according to the standard.

"the argument is an int, the value of which shall be representable as an
unsigned char or shall equal the value of the macro EOF.
If the argument has any other value, the behavior is undefined."

### DIFF
--- a/src/choices.c
+++ b/src/choices.c
@@ -20,18 +20,21 @@ min_match_length(char *str, char *query)
 	size_t mstart;
 	size_t qpos;
 	size_t mpos;
+	int query_char;
 
 	for (mlen = 0, mstart = 0; str[mstart] != '\0'; ++mstart)
-		if (tolower(str[mstart]) == tolower(query[0])) {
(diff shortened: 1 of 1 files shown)


calleerlandsson:
Hey @GBuella, thanks for fixing this!

Does this build on your system when casting to `unsigned char` instead of `unsigned`? If so, I think it would be worth changing to that since that is what the standard is specifying.

GBuella:
All right, I guess that doesn't technically change anything, but might make it more clear for some reading the code.
Added in another commit.

calleerlandsson:
Thanks @GBuella! Merged as cbd0e7a.

PR #55, opened 2015-05-08
Use separate input and output streams

### LINKED ISSUE #45: Typed characters are not read when given a big list of choices and typing fast
Gary Bernhardt was good enough to mention this.

https://twitter.com/garybernhardt/status/575379182684766208

I've not looked into the code, but a possible solution could be a short wait before filtering after a character is typed while you look for more characters, sort of like a buffered filter? Also, threads.

calleerlandsson:
When feeding pick a lot of choices (> 100,000), typing fast on the keyboard
while entering the search query could make pick drop input characters.

By using separate input and output streams rather than one stream for both,
we avoid this issue.

Closes #45

### DIFF
--- a/src/ui.c
+++ b/src/ui.c
@@ -64,7 +64,8 @@ void	 putstrat(int, int, char *);
 void	 move_cursor_to(int, int);
 void	 tty_putp(const char *str);
 
-FILE *tty;
+FILE *tty_out;
+FILE *tty_in;
 struct termios oldattr;
 int using_alternate_screen;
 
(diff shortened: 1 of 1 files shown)


mike-burns:
The code makes sense. That nesting in `get_selected` is deep; perhaps it could be simplified by swapping the `ERR` check and using the fact that err(3) exits. e.g.

```c
if ((c = getc(tty_in)) == ERR)
        err(1, "getc");
if (ch != KEY_BRACKET && ch != KEY_RAW_O)
        break;
if ((ch = getc(tty_in)) == ERR)
        err(1, "getc");
switch (ch) {
...
}
```

calleerlandsson:
Yeah, it is terrible. That's a great refactoring! I'd like to get this PR in, in order to squash this bug but I'd be happy to include that in my coming refactoring PR.

calleerlandsson:
I will merge this.

PR #56, opened 2015-05-08
Sort choices lexicographically

### LINKED ISSUE #50: Does not retain order of results
Steps to reproduce

- `cat file | pick`
- Notice the ordering of the results
- Add some input
- Remove it
- The ordering of the results is not the original anymore

calleerlandsson:
Choices with the same score would be sorted randomly creating an
nondeterministic ordering which confused users.

Choices will now be sorted lexicographically within the same score.

Closes #50

### DIFF
--- a/src/choices.c
+++ b/src/choices.c
@@ -83,7 +83,9 @@ merge(struct choice *front, struct choice *back)
 	c = &head;
 
 	while (front != NULL && back != NULL) {
-		if (front->score > back->score) {
+		if (front->score > back->score ||
+                (front->score == back->score &&
+                 strcmp(front->str, back->str) < 0)) {
 			c->choices.sle_next = front;
 			c = front;
 			front = front->choices.sle_next;


src/choices.c
```diff
@@ -83,7 +83,9 @@ merge(struct choice *front, struct choice *back)
 	c = &head;
 
 	while (front != NULL && back != NULL) {
-		if (front->score > back->score) {
+		if (front->score > back->score ||
+                (front->score == back->score &&
+                 strcmp(front->str, back->str) < 0)) {
```
mike-burns: It might be faster to use strncmp(3). Or maybe that will just lead to bad results. What do you think?

calleerlandsson: I don't know. What would `n` be? The length of the shortest string? Is that what `strcmp` does?

mike-burns: Yeah, I guess that's what strcmp(3) does. If `front->str` and `back->str` are definitely strings, I guess strcmp(3) is faster.

calleerlandsson:
I will merge this.

PR #61, opened 2015-06-22
Feature/use query

### LINKED ISSUE #58: Use input directly
It would be quite useful if `pick` had a bind that used what the user currently has entered.
`dmenu` and some of its clones like `rofi` do this with `Control+Enter` or `Shift+Enter`, though those keys won't work well in the terminal.

This would let users select things that are not in the list.  Consider the following example, which lists local branches in a git repo:

```bash
git branch | cut -c 3- | pick | xargs git checkout
```

With the proposed feature, users could type in a commit hash or a tag, which would not be present in the list but are valid refs to checkout.  Or if a user has a bind like I do to "check out a branch if it exists, and otherwise make it", this would then be possible with `pick` alone.

rosshadden:
/closes #58 

Using `Alt-Enter` in a terminal that outputs an escape prefix will use the user's input query directly.

This lets `pick` be used in situations where the user wishes to enter something not necessarily in the choice list.  For example:

```bash
git checkout $(git branch | cut -c 3- | pick)
```

With this change, users may checkout a `git` hash or tag, along with the original behavior of selecting a branch from a list.

### DIFF
--- a/man/man1/pick.1
+++ b/man/man1/pick.1
@@ -56,6 +56,8 @@ interface is operated with the following keys:
 Select between choices matching the current search query.
 .It Ic Enter
 Output the currently selected choice and exit.
+.It Ic Alt\&-Enter
+Output the current input query and exit.
 .It Ic "Printable characters"
 Printable characters are added to the search query input field and will refine
 the current search query.
--- a/src/tty.c
+++ b/src/tty.c
@@ -86,6 +86,10 @@ tty_getch()
 	if (ch == ESCAPE) {
 		ch = tty_getc();
 
+		if (ch == '\n') {
+			return TTY_ALT_ENTER;
+		}
+
 		if (ch == '[' || ch == 'O') {
 			ch = tty_getc();
 
--- a/src/tty.h
+++ b/src/tty.h
@@ -13,6 +13,7 @@
 #define TTY_CTRL_W 23
 #define TTY_DEL 127
 #define TTY_ENTER 10
+#define TTY_ALT_ENTER 266
 #define TTY_BACKSPACE 263
 #define TTY_UP 259
 #define TTY_DOWN 258
--- a/src/ui.c
+++ b/src/ui.c
@@ -74,6 +74,8 @@ ui_selected_choice(struct choices *choices, char *initial_query,
 				free(query);
 				return selected_choice(choices, selection);
 			}
+
+			break;
 		case TTY_ALT_ENTER:
 			tty_restore();
 			return choice_new(query, "", 1);
--- a/src/ui.c
+++ b/src/ui.c
@@ -74,6 +74,9 @@ ui_selected_choice(struct choices *choices, char *initial_query,
 				free(query);
 				return selected_choice(choices, selection);
 			}
+		case TTY_ALT_ENTER:
+			tty_restore();
+			return choice_new(query, "", 1);
 		case TTY_CTRL_N:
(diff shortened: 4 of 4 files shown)


rosshadden:
@calleerlandsson Assuming you like the feature addition, there's a decision we need to make before it gets merged in.

There is currently no `break;` statement at the end of the [`TTY_ENTER` case](https://github.com/rosshadden/pick/blob/d3bca6d123805a2314f2abfdfa13f768e2940b7e/src/ui.c#L71-L76).  What this means is that if the user's query has 0 `visible_choices_count`, merely pressing `Enter` (without `Alt`) will use the input query, as execution falls into my `TTY_ALT_ENTER` case.

Said more plainly, it means that typing `blah<Enter>` when there are no matches will output `blah`.

I actually like this behavior a lot, but I'm leaving it up to you to decide if it should be in `pick`.  If you don't, I merely add a `break;` to the end of the preceding case.

calleerlandsson:
Awesome @rosshadden! Thanks for working on this.

I do think we should add a `break` statement to the `TTY_ENTER` case since not doing so would change existing behavior. If it becomes obvious that the `Alt+Enter` behavior is the one we use most of the time, I'm totally open to change this in the next major release.

If I get time to look at the code during the week, I will. Otherwise, I'll look at it on Friday.

rosshadden:
Okay cool.  I agree, this way it's consistent.  I made the change and this is ready to be looked over.  No rush---I'm already using my fork, and I daresay it has greatly improved my use cases of `pick` :dog: 

Thanks!

calleerlandsson:
Thanks @rosshadden!

I've checked out the code, tried it out locally, and it works great!

The only thing that came to mind when looking at the code was that we're not freeing some memory on exit. More specifically, the query string and the new choice are not freed.

Since we're exiting, it's not that big a deal but I think making sure to free memory will make the program easier to change and refactor in the future.

What do you think about this patch?

```patch
diff --git a/src/ui.c b/src/ui.c
index 88c5abc..04eaa44 100644
--- a/src/ui.c
+++ b/src/ui.c
@@ -40,6 +40,7 @@ ui_selected_choice(struct choices *choices, char *initial_query,
        char *query;
        int key, selection, visible_choices_count, word_position;
        size_t cursor_position, query_size, query_length, initial_query_length;
+       struct choice *choice;
 
        initial_query_length = strlen(initial_query);
        cursor_position = initial_query_length;
@@ -78,7 +79,10 @@ ui_selected_choice(struct choices *choices, char *initial_query,
                        break;
                case TTY_ALT_ENTER:
                        tty_restore();
-                       return choice_new(query, "", 1);
+                       choice = choice_new(query, "", 1);
+                       SLIST_INSERT_HEAD(choices, choice, choices);
+                       free(query);
+                       return choice;
                case TTY_CTRL_N:
                        if (selection < visible_choices_count - 1) {
                                ++selection;
```

Since `query` will be `strdup`ed in `choice_new`, we're free to `free` it after calling `choice_new`. Adding the choice to the list of choices before exiting will allow it to be freed by the call to `choices_free` on https://github.com/thoughtbot/pick/blob/d2831c6926f50677e61b135da78a3a3d946e059c/src/main.c#L64.

rosshadden:
Yeah, it looks good to me.  Thanks, I don't write C often.  But when I do... etc. etc.

calleerlandsson:
Cool! Then I'll apply it before merging this on Friday. Thanks!

PR #63, opened 2015-06-27
Levenshtein substring matching

GBuella:
A functioning version of edit distance based matching.

Can switch to this other matching mode using ctrl+x
Filters strings which have a substring with edit distance lower than three from the query string

Not complete change, still lacking:
mention in man page
configurable edit distance treshold
handle any long strings
tuning for speed - slow with tens of thousands of strings, but can be coded to run more efficiently


calleerlandsson:
Hey @GBuella, thanks so much for contributing!

I'm not familiar with and have yet to try Levenshtein edit distance based search.

Searching through the same strings, how does it differ from the current algorithm in pick? Is it better for searching through a specific type of strings? Do you think it could replace the current algorithm?

If I get time during the week, I'll take a look at the code. If not, I'll look at it on Friday.

GBuella:
Hey @calleerlandsson , I attempt to solve two things with this: matching typos, and excluding a lot of unnecessary noise in the results.

Looking for substrings, instead of just characters in the right order

consider a user typing "makefile", the following inputs do match:

~/something/makefile
~/ma/teker/alf/miles_per_hour  < due to "ma" + "ke" + "f" + "ile"

Often I totally don't need the second of these.
A partial solution would to only look for exactly matching substring, but that is still not good enough:

Looking for "makefile":

~/something/makefile  <- found
~/ma/teker/alf/miles_per_hour  <- not found
~/something/other/makfile  <- typo, not found

I would like to find the substring "makfile" as well, which has an edit distance of 1 form the original query "makefile"

Thus looking for substrings with sufficiently low distance from the query:

E.g.: looking for "some/makefile" finds:

~/code/some//makefile       <- not two slashes
~/code/other/some//makfile    <- missing "e" in "makfile"
some makefile    <-   no slash, but a match anyways
some/makelife    <-  typo,  makelife vs. makefile

and similar matching strings, while excluding a lot of unnecessary long input lines, that just happen to have the characters in the right order.

If find some more time, I'll try to write a more clear explanation. For now, I hope this explanation is understandable.

calleerlandsson:
Hey @GBuella,

This is a cool algorithm and seems to fit your use-case very well!

I'm not sure this feature would be a good fit for merging into pick.

Here are some reasons to why I'm hesitant to do so:

- The way the current algorithm works is a big part of what pick is. The
  ability to type `scacsp<CR>` to jump straight to
  `spec/controllers/application_controller_spec.rb` is specific to the
  algorithm and in line with what I think people generally expect of a
  fuzzy finder.

- The feature is adding a considerable amount of code that will have to
  be debugged, maintained and considered during refactoring and
  development.

Since I'm using pick differently, I won't use the feature provided by
this code and will therefore struggle with maintaining the it.

I want to make sure you know that I'm very thankful for the time you've
spent on this and that I can clearly see how this supports your
use-case.

I'm trying to think of a plugin architecture that would allow for anyone
to provide their own search algorithm but have yet to come up with
anything good that allows pick to continue to be performant even with
large sets of choices.

Do you think a plugin architecture would make sense? Do you have any
ideas for how one could work? Would using this algorithm as a plugin fit
your use-case?

I'd be very happy to hear your ideas about such an architecture!

GBuella:
Hi, in response:

"Since I'm using pick differently, I won't use the feature provided"  that is why the default behaviour is kept, and this one is optinal ( currently enabled by pressing ctrl-X )

"will have tobe debugged, maintained and considered during refactoring "  True, although I can surely help with that

"Do you think a plugin architecture would make sense"  I don't know for sure, I have no idea how many other devs would like to make plugins.
If it is easy enough to make a plugin, perhaps some poeple would make plugins. As of now, if this one would be the only plugin, not much use of a plugin architecture -- it would still mean maintaining a few hundred lines of code only for this.

Anyways, I'm just going to use this from my own repo then, so it doesn't matter much to me either way... and if one or two other devs would like another special way of matching, then a plugin system surely makes a lot of sense.

Cheers!

calleerlandsson:
Thanks for understanding, @GBuella.

I'll think more about the plugin infrastructure. If anyone has ideas or wants to submit a PR for that, I'd very much appreciate it!

I will close this PR in the meantime.

PR #64, opened 2015-07-03
Prefer `reallocarray` over `realloc`

### LINKED ISSUE #59: Prevent potential multiplication overflow when allocating memory
When allocating memory using `malloc` there is a risk for multiplication overflow if the call looks anything like `malloc(sizeof(whatever) * n)`. To prevent this `calloc(sizeof(whatever), n)` can be used instead.

Similarly, calls like `realloc(sizeof(whatever) * n)` suffers from the same problems and can be replaced with calls to `reallocarray`. `reallocarray` is implemented in OpenBSDs stdlib and can be added to `compat/` for portability.

calleerlandsson:
When reallocating memory using `realloc`, passing in a number produced by a
multiplication operation, there is a risk of multiplication overflow:

    realloc(str, len * sizeof(char))`

These overflow issues can be avoided by using `reallocarray`.

`reallocarray` is implemented in the OpenBSD libc and has been added to
`compat/` for portability.

Closes #59.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -29,6 +29,9 @@
    and to 0 otherwise. */
 #undef HAVE_REALLOC
 
+/* Define to 1 if you have the `reallocarray' function. */
+#undef HAVE_REALLOCARRAY
+
 /* Define to 1 if you have the `setlocale' function. */
 #undef HAVE_SETLOCALE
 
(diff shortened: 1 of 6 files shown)


mike-burns:
LGTM.

calleerlandsson:
Thanks @mike-burns! Merging.

PR #67, opened 2015-07-19
Flush TTY when a full screen is drawn

ScoreUnder:
The "pick" utility was not behaving well for me at all, and was unusable in both xterm and urxvt. I don't know why my tty is so different, but this fixes things for me.

Oddly enough the 1.1.1 release worked fine for me.

### DIFF
--- a/src/tty.c
+++ b/src/tty.c
@@ -164,6 +164,12 @@ tty_columns()
 	return columns;
 }
 
+int
+tty_flush()
+{
+	return fflush(tty_out);
+}
+
 static int
 tty_getc()
 {
--- a/src/tty.h
+++ b/src/tty.h
@@ -31,5 +31,6 @@ void tty_exit_standout_mode();
 void tty_move_cursor_to(int, int);
 int tty_lines();
 int tty_columns();
+int tty_flush();
 
 #endif /* TTY_H */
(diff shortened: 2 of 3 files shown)


calleerlandsson:
Hey @ScoreUnder, thanks for contributing!

I have been traveling and am sorry you haven't heard back from me until now.

This change looks good. I'll try it out this Friday and get back to you!

calleerlandsson:
Hey @ScoreUnder, I tried you branch out locally and it works great!

While I was trying it, I realized that flushing should only be required if the output is buffered. Since we want output to show up at the tty directly, I figured we could try setting the output stream to be unbuffered instead of flushing it on every iteration.

I have pushed a branch called [unbuffered-out](https://github.com/thoughtbot/pick/tree/unbuffered-out) that does just this. Could you try it locally to see if it fixes the issues you were having?

ScoreUnder:
The unbuffered-out branch works, but compare:

Unbuffered:
Screenshot
(Note, I'm not sure why there are white flashes here - I couldn't see them while actually using the program)

Buffered with flush:
Screenshot

The unbuffered version sometimes draws a half-done frame before it finishes, and while you type or press the arrow keys, the cursor noticeably flickers. It's still a great improvement over the previous behaviour I was seeing.

Current master:
Screenshot

calleerlandsson:
Thanks for trying it out! It definitely seems like the flushing buffered versions works best so lets go with that one. I'll merge this.

Issue #69, opened 2015-08-09
pick -d splits by the first occurrence of IFS


temochka:
The man page currently states the following:

> When the -d option is supplied, input lines will be split into two parts by the **last occurrence of IFS**.

However, the observed behavior is the opposite of that (split by the first occurrence):

```
$ pick -d <<< "foo bar baz qux"
foo
```

I’ll be happy to help modify `io_read_choices` to match the documented behavior, if you’d like to fix that in the code.

P.S. There is also a typo in word "occurrence" in this man page.

calleerlandsson:
Hey @temochka, thanks for reporting this! This is definitely a bug and a PR would be much welcome!

temochka:
Sounds good. I’ll try to work on it later this week!

calleerlandsson:
Thanks!

PR #70, opened 2015-08-12
Fix pick -d splitting behavior

temochka:
> When the -d option is supplied, input lines will be split into two parts by the last occurrence of IFS.

See #69 for more details. Replaced `strtok_r` with eager `strpbrk` to avoid unnecessary string mutations. Also fixed a typo in documentation. Tested on OS X 10.10.4.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -20,15 +20,15 @@ static char * eager_strpbrk(const char *, const char *);
 struct choices *
 io_read_choices(int read_descriptions)
 {
-	char *line, *description, *field_separator;
+	char *line, *description, *field_separators;
 	size_t line_size;
 	ssize_t length;
 	struct choice *choice;
 	struct choices *choices;
 
-	field_separator = getenv("IFS");
-	if (field_separator == NULL) {
-		field_separator = " ";
+	field_separators = getenv("IFS");
+	if (field_separators == NULL) {
+		field_separators = " ";
 	}
 
 	choices = malloc(sizeof(struct choices));
@@ -49,7 +49,7 @@ io_read_choices(int read_descriptions)
 
 		chomp(line, length);
 
-		if (read_descriptions && (description = eager_strpbrk(line, field_separator))) {
+		if (read_descriptions && (description = eager_strpbrk(line, field_separators))) {
 			*description++ = '\0';
 		} else {
 			description = "";
--- a/src/io.c
+++ b/src/io.c
@@ -15,6 +15,7 @@
 #include "choices.h"
 
 static void chomp(char *, ssize_t);
+static char * eager_strpbrk(const char *, const char *);
 
 struct choices *
 io_read_choices(int read_descriptions)
@@ -39,7 +40,6 @@ io_read_choices(int read_descriptions)
 
 	for (;;) {
 		line = NULL;
-		description = NULL;
 		line_size = 0;
 
 		length = getline(&line, &line_size, stdin);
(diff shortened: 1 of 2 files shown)


src/io.c
```diff
@@ -85,3 +83,14 @@ chomp(char *string, ssize_t length)
 		string[length - 1] = '\0';
 	}
 }
+
+static char *
+eager_strpbrk(const char *string, const char *separators) {
+	char *ptr = NULL, *tmp_ptr;
+	for (tmp_ptr = strpbrk(string, separators);
+	     tmp_ptr;
+	     tmp_ptr = strpbrk(tmp_ptr, separators)) {
```
calleerlandsson: Is this the same as the following and would you in that case think the following is clearer?

```c
tmp_ptr = strpbrk(string, separators);
while (tmp_ptr = strpbrk(tmp_ptr, separators)) {
```

src/io.c
```diff
@@ -85,3 +83,14 @@ chomp(char *string, ssize_t length)
 		string[length - 1] = '\0';
 	}
 }
+
+static char *
+eager_strpbrk(const char *string, const char *separators) {
```
calleerlandsson: Should `separators` be singular `separator`?

calleerlandsson:
Hey @temochka, thanks for submitting this PR! I had two questions, other than that, this looks great!

src/io.c
temochka: Having done far more C++ than C I’m definitely biased towards `for` loops. However, in this particular case I believe that `for` loop actually improves readability. Your suggested implementation is incomplete, since `strpbrk` can return `NULL` on the first call, meaning we’d need to handle this special case with an extra `if` statement. Like this:

``` c
static char *
eager_strpbrk(const char *string, const char *separators) {
	char *ptr, *tmp_ptr;
	ptr = tmp_ptr = strpbrk(string, separators);

	if (ptr == NULL) {
		return NULL;
	}

	while ((tmp_ptr = strpbrk(++tmp_ptr, separators))) {
		ptr = tmp_ptr;
	}
	
	return ptr;
}
```

On the contrary, `for` loop handles NULL pointers naturally. But if you still think that `while`-based implementation is clearer or more consistent with the rest of the project’s codebase, I don’t mind changing my commit to it.

src/io.c
temochka: Given that `$IFS` is actually a list of possible field separating characters, I think `separators` is a better name for what it is. In case you weren’t aware of this behavior, this is exactly how `strtok_r` works too, so there should be no regression. `strtok_r` [uses](https://github.molgen.mpg.de/git-mirror/glibc/blob/master/string/strtok_r.c#L63) `strpbrk` internally, while `strpbrk` is [implemented via two loops](https://github.molgen.mpg.de/git-mirror/glibc/blob/master/string/strpbrk.c) with the outer loop iterating a given string char-by-char and the inner one checking these characters for inclusion in separators.

Default `$IFS` set by Bash includes all regular whitespace characters (space, tab and newline):

``` bash
$ echo "$IFS" | cat -te
 ^I$
$
```

calleerlandsson: That makes a lot of sense. Thanks for explaining it! Should we rename `field_separator` to `field_separators` or `separators` for parity?

src/io.c
calleerlandsson: That is a very good point! Thanks for explaining. It's obvious that the `for` loop is superior in this case!

src/io.c
temochka: Makes sense. Pushed another commit that does exactly that.

calleerlandsson:
Thanks @temochka! I will get time to merge this in on Friday!

temochka:
Sure, take your time, and thanks for the heads-up!

calleerlandsson:
Merged as 5e243a5b1666573a349f33911c3d22779eac0899, 3259fadac76914857f16db5b048e651bcf7e7c8b, and 011075fb249f97fdee2df784eda54d0b857b45b3. Thanks, @temochka!

temochka:
No problem, I was glad to help. Thank you too, @calleerlandsson!

PR #74, opened 2015-10-16
Prevent scrolling

calleerlandsson:
When displaying as many or more results than the number of lines in the
terminal, when the last result's length was longer than the width of the
terminal, the extra characters would push the first result up into the
input field.

By not printing more characters than would fit the terminal width, this
behavior is avoided.


mike-burns:
Simple solution. LGTM.

calleerlandsson:
Thanks, @mike-burns!

PR #76, opened 2015-10-29
README.md: fix incorrect filename; INSTALL.md.in: Add missing instruc…

FreedomBen:
…tion

In README.md the filename of the INSTALL file containing instructions is
incorrect.  It should be INSTALL.md.in (Fixed in this commit)

In INSTALL.md.in there is a missing step.  You first should run
./autogen.sh to generate the configure file

### DIFF
--- a/INSTALL.md.in
+++ b/INSTALL.md.in
@@ -9,6 +9,7 @@ pick installation
 
 2. Configure the distribution.
 
+        ./autogen.sh
         ./configure
 
 3. Build and install pick:
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ tar -xzf pick-VERSION.tar.gz
 
 ```sh
 cd pick-VERSION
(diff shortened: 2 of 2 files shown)


calleerlandsson:
Hey @FreedomBen, thanks for contributing!

Did you download the latest release? When creating the tarball that's uploaded and published for every release Automake will create `INSTALL.md` from `INSTALL.md.in`, it will also run the equivalent of `./autogen.sh` so that when you unpack the tarball you can run `./configure` directly.

If you want to build directly from `master` you should follow the instructions in `DEVELOPING.md`. Does these seem more correct to you?

FreedomBen:
Oh yeah, that does make sense.  I never downloaded the tarball so that explains the confusion.  Thanks!  And pick is super awesome by the way.

calleerlandsson:
Thanks @FreedomBen, I'm glad you like it!

PR #77, opened 2015-11-08
Consolidate files

calleerlandsson:
This PR consolidates all source files into `src/main.c`, all compat files into `src/compat.h` and `src/compat.c`, and all automake files into `Makefile.am`.

Doing this makes the pick source code easier to read and follow since functions aren't scattered across many files and allows for healthy refactoring.


mike-burns:
That is much simpler, and seems like a good base from which we can refactor and extract more. LGTM.

calleerlandsson:
Thanks @mike-burns! I'll merge this.

PR #79, opened 2015-12-05
Don't clear the screen

mptre:
The screen should not be cleared unless the alternate screen is being
used. This change requires all cursor movements to be relative rather
than absolute. By initially emitting the number of choices and saving
the cursor the cursor can always be restored at its initial position.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -78,10 +78,8 @@ static void		 tty_putp(const char *);
 static int		 tty_putc(int);
 static void		 handle_sigint(int);
 static void		 restore_tty(void);
-static void		 put_line(int, char *, int, int);
+static void		 put_line(char *, int, int);
 static void		 start_standout(void);
-static void		 print_at(int, int, char *, int);
-static void		 move_to(int, int);
 static void		 end_standout(void);
 static int		 print_choices(int);
 static void		 show_cursor(void);
@@ -91,6 +89,7 @@ static struct choice	*choice_at(int);
 static void		 delete_between(char *, size_t, size_t, size_t);
 static void		 hide_cursor(void);
 static void		 free_choices(void);
+static void		 print_query(char *, size_t, int);
 
 static FILE		*tty_in;
 static FILE		*tty_out;
@@ -280,9 +279,8 @@ selected_choice(void)
 
 	init_tty();
 
-	put_line(0, query, query_length, 0);
 	visible_choices_count = print_choices(selection);
-	move_to(0, cursor_position);
+	print_query(query, query_length, cursor_position);
 	show_cursor();
 
 	for (;;) {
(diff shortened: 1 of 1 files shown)


calleerlandsson:
Thanks for submitting this PR, @mptre! I'll take a closer look at this and get back to you when I've had time to do so!

src/main.c
```diff
@@ -591,7 +591,16 @@ init_tty(void)
 	if (use_alternate_screen)
 		tty_putp(enter_ca_mode);
 
-	tty_putp(clear_screen);
+	/* Emit enough lines to fit all choices. */
```
calleerlandsson: What is the reason for emitting these lines?

src/main.c
```diff
@@ -657,38 +668,35 @@ start_standout(void)
...
 print_choices(int selection)
 {
 	char		*line;
-	int		 i, visible_choices_count = 0;
+	int		 visible_choices_count = 0;
 	size_t		 length, line_length = 64;
 	struct choice	*choice;
 
+	/* Emit query line. */
+	tty_putc('\n');
```
calleerlandsson: Could this be avoided by printing the query before printing the choices?

calleerlandsson:
Hey @mptre, I tried your branch locally and it works great. I'm not sure I see any difference in behavior to what is currently on master, though. How can I test this?

src/main.c
mptre: It's pre-allocating the number of lines needed to output the query line and all
input lines that fits on the screen in order to determine the position of the
first line. This could probably be done by moving the cursor up the same amount
of times as the number of matching choices after printing the choices but this
requires more state and I prefer the single save and restore capability.

src/main.c
mptre: Probably but you then need to move the cursor to the right position on query
line after printing the matching choices.

mptre:
Position your prompt at the bottom of the terminal: either by running a command
outputting many lines or by hitting the return key repeatedly. Then run:

  printf 'foo\nbar\n' | pick -X

The pick interface should be drawn at the bottom of the terminal. 

The `-X' isn't necessarily needed but just used as an example to surely disable
the alternate screen.

calleerlandsson:
Following your instructions, on Mac OS X 10.11.2 in Terminal.app, I'm seeing the same behavior on both your branch and master. I'm not sure if I'm seeing the correct behavior or the incorrect behavior.

> The pick interface should be drawn at the bottom of the terminal.

I think this is happening but I'm not sure. Here's an animated gif of what I'm seeing:

pick

Are you seeing something different? If so, could you describe or record it? What OS and terminal emulator are you running?

mptre:
Notice I'm using screen with the alternate screen disabled:

pick

calleerlandsson:
I see! I was using the wrong executable and wasn't seeing the behavior your gif demonstrates.

This looks good to me. I'll merge this.

calleerlandsson:
Merged as dc0c39dc337ffae9a88354f9c2c3e2d1809868f4. Thanks, @mptre!

PR #80, opened 2015-12-14
Improve tty_putp error message

mptre:
When tty_putp fails the following message is outputted:
```
pick: tputs: Success
````
Since tputs does not set errno this message is confusing. Instead output
the name of the missing capability:
````
pick: enter_ca_mode: unknown terminfo capability
````
Notice that enter_ca_mode is just an example and will be substituted with
the capability that caused the failure during execution.

Discovered using pick with the Linux console.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -51,6 +51,11 @@
 #define EX_SIG 128
 #define EX_SIGINT (EX_SIG + SIGINT)
 
+#define tty_putp(capability) do { \
+	if (tputs(capability, 1, tty_putc) == ERR) \
+		errx(1, #capability ": unknown terminfo capability"); \
+	} while (0)
+
 struct choice {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/main.c [hunk 1]
```diff
@@ -51,6 +51,11 @@
 #define EX_SIG 128
 #define EX_SIGINT (EX_SIG + SIGINT)
 
+#define tty_putp(capability) do { \
+	if (tputs(capability, 1, tty_putc) == ERR) \
+		errx(1, #capability ": unknown terminfo capability"); \
```
calleerlandsson: Today I learned about [stringification](https://gcc.gnu.org/onlinedocs/cpp/Stringification.html)! Thanks!

calleerlandsson:
Thanks for submitting this PR, @mptre, it looks great! I'll merge this.

PR #81, opened 2015-12-21
Preserve order of input

mptre:
One annoying aspect of pick is that the order of input lines is not
preserved when the query is empty. By allocating the input in one
consecutive block of memory the fallback sorting method can be
simplified to comparing the address of the two choices corresponding
lines. The initial order of the input will then always be preserved
since &line[i] < &line[i + 1] holds true.

The input allocation and choice creation is done in two separate steps
in order to avoid the need to update any existing previous choice line
pointers if the input memory block gets reallocated and moved to another
region. In terms of theoretical time complexity it's still a linear
operation.

As a pleasant side-effect the number of necessary allocations has been
reduces since the choice string and description is no longer duplicated.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -203,7 +203,7 @@ get_choices(void)
 	if ((input.string = malloc(input.size)) == NULL)
 		err(1, "malloc");
 	for (;;) {
-		if ((length = read(0, input.string + input.length,
+		if ((length = read(STDIN_FILENO, input.string + input.length,
 				   input.size - input.length)) <= 0)
 		    break;
 
--- a/src/main.c
+++ b/src/main.c
@@ -105,8 +105,8 @@ static struct termios	 original_attributes;
 static struct choices	*choices;
 static struct {
 	size_t	size;
-	size_t	nmemb;
-	char	*v;
+	size_t	length;
+	char	*string;
 } input;
 
 int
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -104,6 +103,11 @@ static int		 use_alternate_screen;
 static size_t		 query_size;
 static struct termios	 original_attributes;
 static struct choices	*choices;
+static struct {
+	size_t	size;
```
mptre: The struct members should probably be renamed in order to match the current naming convention of pick.

src/main.c
```diff
@@ -188,48 +192,50 @@ version(void)
...
 	struct choice	*choice;
 
 	if ((field_separators = getenv("IFS")) == NULL)
 		field_separators = " ";
 
+	input.size = BUFSIZ;
+	if ((input.v = malloc(input.size)) == NULL)
+		err(1, "malloc");
+	for (;;) {
+		if ((length = read(0, input.v + input.nmemb,
```
calleerlandsson: Do you think replacing `0` with `STDIN_FILENO` would be make this easier to read?

src/main.c
calleerlandsson: What do you think about the following member names?

```
size_t size;
size_t length;
char *string;
```

calleerlandsson:
Hey @mptre, this looks great! Thanks for contributing! I commented with a couple of questions above.

Since this is a change in behavior, do you think we need to increment the major or minor version number when releasing?

src/main.c
mptre: Fixed

src/main.c
mptre: Fixed. Not a fan but it feels more consistent with the current codebase.

On Tue, Dec 22, 2015 at 01:27:09AM -0800, Calle Erlandsson wrote:
> >  	ssize_t		 length;
> >  	struct choice	*choice;
> >  
> >  	if ((field_separators = getenv("IFS")) == NULL)
> >  		field_separators = " ";
> >  
> > +	input.size = BUFSIZ;
> > +	if ((input.v = malloc(input.size)) == NULL)
> > +		err(1, "malloc");
> > +	for (;;) {
> > +		if ((length = read(0, input.v + input.nmemb,
> 
> Do you think replacing `0` with `STDIN_FILENO` would be make this easier to read?
> 
> ---
> https://github.com/thoughtbot/pick/pull/81/files#r48235527

-- 
:wq

mptre:
Good question. Since it a behavioural change it might be more suitable with a
major release for easier distinction between the prior and later behavior. On
the other hand I've always considered the prior ordering a bug and therefore a
minor would be sufficient :)

On Tue, Dec 22, 2015 at 01:32:43AM -0800, Calle Erlandsson wrote:
> Hey @mptre, this looks great! Thanks for contributing! I commented with a couple of questions above.
> 
> Since this is a change in behavior, do you think we need to increment the major or minor version number when releasing?
> 
> ---
> https://github.com/thoughtbot/pick/pull/81#issuecomment-166566145

-- 
:wq

calleerlandsson:
Thanks, @mptre. I'll merge this.

calleerlandsson:
Merged as bf58c61a7639ad9a45cc33e9d47488367fed7ae1.

PR #82, opened 2015-12-29
Replace linked list with array

mptre:
This is a substantial change replacing the usage of a linked list as the
internal data structure for the choices with an array. Here are some
advantages in no particular order:

- Arrays are simple and have a great cache locality

- A large portion of the code in `src/compat.h` can be removed

- Enables usage of the libc `qsort(3)` sort function

- Reduces LOC

Still left TODO:

  - Rename `choices` members in order to adapt the convention of codebase

  - Remove all SLIST/QUEUE related code from `src/compat.h` and automake

On a side note: AFAIK setting the choice score to `1` is used in order to trigger sorting when the query is empty. Therefore `1` should be defined as a macro with a more descriptive name. While at it: the `0` score should also be defined as a macro.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -652,7 +652,7 @@ print_choices(int selection)
 
 	for (i = 0;
 	     i < (ssize_t)choices.nmemb
-	     && i < lines
+	     && i < lines - 1
 	     && (choice = &choices.v[i])->score != 0;
 	     ++i) {
 		length = strlen(choice->string) + strlen(choice->description) +
--- a/src/main.c
+++ b/src/main.c
@@ -57,26 +57,20 @@
 	} while (0)
 
 struct choice {
-	SLIST_ENTRY(choice)	 choices;
-	char			*description;
-	char			*string;
-	float			 score;
+	char	*description;
+	char	*string;
+	float	 score;
 };
-
-SLIST_HEAD(choices, choice);
 
 __dead static void	 usage(void);
 __dead static void	 version(void);
 static void 		 get_choices(void);
 static char		*eager_strpbrk(const char *, const char *);
-static struct choice	*new_choice(char *, char *, float);
 static void		 put_choice(struct choice *);
 static struct choice	*selected_choice(void);
 static void		 filter_choices(void);
 static float		 score(char *);
 static size_t		 min_match_length(char *);
-static struct choice	*sort(struct choice *);
-static struct choice	*merge(struct choice *, struct choice *);
 static void		 init_tty(void);
 static int		 tty_putc(int);
 static void		 handle_sigint(int);
@@ -88,11 +82,11 @@ static int		 print_choices(int);
 static void		 show_cursor(void);
 static int		 get_key(void);
 static int		 tty_getc(void);
-static struct choice	*choice_at(int);
 static void		 delete_between(char *, size_t, size_t, size_t);
 static void		 hide_cursor(void);
 static void		 free_choices(void);
 static void		 print_query(char *, size_t, int);
+static int		 sort_choices(const void *, const void *);
 
 static FILE		*tty_in;
 static FILE		*tty_out;
@@ -102,7 +96,11 @@ static int		 output_description = 0;
 static int		 use_alternate_screen;
 static size_t		 query_size;
 static struct termios	 original_attributes;
-static struct choices	*choices;
+static struct {
+	size_t size;
+	size_t nmemb;
+	struct choice *v;
+} choices;
 static struct {
 	size_t	size;
 	size_t	length;
@@ -194,7 +192,6 @@ get_choices(void)
 {
 	char		*description, *field_separators, *start, *stop;
 	ssize_t		 length;
-	struct choice	*choice;
 
 	if ((field_separators = getenv("IFS")) == NULL)
 		field_separators = " ";
@@ -216,10 +213,9 @@ get_choices(void)
 	}
 	memset(input.string + input.length, '\0', input.size - input.length);
 
-	if ((choices = malloc(sizeof(struct choices))) == NULL)
+	choices.size = 16;
+	if ((choices.v = malloc(choices.size*sizeof(struct choice))) == NULL)
 		err(1, "malloc");
-
-	SLIST_INIT(choices);
 
 	start = input.string;
 	while ((stop = strchr(start, '\n')) != NULL) {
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -703,7 +650,11 @@ print_choices(int selection)
 	if ((line = calloc(sizeof(*line), line_length)) == NULL)
 		err(1, "calloc");
 
-	SLIST_FOREACH(choice, choices, choices) {
+	for (i = 0;
+	     i < (ssize_t)choices.nmemb
+	     && i < lines
```
calleerlandsson: I think there might be an off by one error here. Running pick with more choices than would fit on screen, the selection is invisible. I think this is because the last choice printed to the terminal makes it scroll and then the first choice is overwritten by the query line.

Changing this to `&& i < lines + 1` locally fixes the issue for me.

calleerlandsson: Sorry, `- 1` that is.

src/main.c
```diff
@@ -602,13 +553,9 @@ init_tty(void)
 		tty_putp(enter_ca_mode);
 
 	/* Emit enough lines to fit all choices. */
-	SLIST_FOREACH(choice, choices, choices) {
-		choices_count++;
+	for (i = 0; i < (ssize_t)choices.nmemb && i < lines; ++i)
```
calleerlandsson: Given https://github.com/thoughtbot/pick/pull/82/files#r48715231 is correct, should it be `lines - 1` here too?

src/main.c
```diff
@@ -102,7 +96,11 @@ static int		 output_description = 0;
 static int		 use_alternate_screen;
 static size_t		 query_size;
 static struct termios	 original_attributes;
-static struct choices	*choices;
+static struct {
+	size_t size;
+	size_t nmemb;
```
calleerlandsson: What do you think about naming this length to be more consistent with https://github.com/thoughtbot/pick/pull/82/files#diff-803c5170888b8642f2a97e5e9423d399R106?

src/main.c
```diff
@@ -460,13 +451,25 @@ selected_choice(void)
...
-		choice->score = score(choice->string);
-	}
+	for (i = 0; i < (ssize_t)choices.nmemb; ++i)
+		choices.v[i].score = score(choices.v[i].string);
+
+	qsort(choices.v, choices.nmemb, sizeof(struct choice), sort_choices);
+}
+
+static int
+sort_choices(const void *p1, const void *p2)
```
calleerlandsson: What do you think about naming this something like `compare_choices` or `cmp_choices` to denote that it serves as a comparison function?

calleerlandsson:
Hey @mptre, this is awesome! Thanks for your great work! I left some comments above but other than that this looks great and I'd be happy to merge it. I also agree with your `TODO` notes.

src/main.c
mptre: Great catch, fixed.

On Mon, Jan 04, 2016 at 01:36:56AM -0800, Calle Erlandsson wrote:
> > @@ -703,7 +650,11 @@ print_choices(int selection)
> >  	if ((line = calloc(sizeof(*line), line_length)) == NULL)
> >  		err(1, "calloc");
> >  
> > -	SLIST_FOREACH(choice, choices, choices) {
> > +	for (i = 0;
> > +	     i < (ssize_t)choices.nmemb
> > +	     && i < lines
> 
> I think there might be an off by one error here. Running pick with more choices than would fit on screen, the selection is invisible. I think this is because the last choice printed to the terminal makes it scroll and then the first choice is overwritten by the query line.
> 
> Changing this to `&& i < lines + 1` locally fixes the issue for me.
> 
> ---
> https://github.com/thoughtbot/pick/pull/82/files#r48715092

-- 
:wq

src/main.c
mptre: In this case there is no off-by-one error since starting the program will
place the cursor on the next line after the prompt (i.e. the query line) and
therefore `min(choices.nmemb, lines)` lines needs to be emitted.

On Mon, Jan 04, 2016 at 01:40:33AM -0800, Calle Erlandsson wrote:
> > @@ -602,13 +553,9 @@ init_tty(void)
> >  		tty_putp(enter_ca_mode);
> >  
> >  	/* Emit enough lines to fit all choices. */
> > -	SLIST_FOREACH(choice, choices, choices) {
> > -		choices_count++;
> > +	for (i = 0; i < (ssize_t)choices.nmemb && i < lines; ++i)
> 
> Given https://github.com/thoughtbot/pick/pull/82/files#r48715231 is correct, should it be `lines - 1` here too?
> 
> ---
> https://github.com/thoughtbot/pick/pull/82/files#r48715292

-- 
:wq

src/main.c
mptre: Fixed, how about `v`?

On Mon, Jan 04, 2016 at 01:42:46AM -0800, Calle Erlandsson wrote:
> > @@ -102,7 +96,11 @@ static int		 output_description = 0;
> >  static int		 use_alternate_screen;
> >  static size_t		 query_size;
> >  static struct termios	 original_attributes;
> > -static struct choices	*choices;
> > +static struct {
> > +	size_t size;
> > +	size_t nmemb;
> 
> What do you think about naming this length to be more consistent with https://github.com/thoughtbot/pick/pull/82/files#diff-803c5170888b8642f2a97e5e9423d399R106?
> 
> ---
> https://github.com/thoughtbot/pick/pull/82/files#r48715414

-- 
:wq

src/main.c
mptre: Makes sense, how about `choicecmp` or `choice_cmp`? Both inspired by `strcmp`.

On Mon, Jan 04, 2016 at 01:44:03AM -0800, Calle Erlandsson wrote:
> >  
> > -	SLIST_FOREACH(choice, choices, choices) {
> > -		choice->score = score(choice->string);
> > -	}
> > +	for (i = 0; i < (ssize_t)choices.nmemb; ++i)
> > +		choices.v[i].score = score(choices.v[i].string);
> > +
> > +	qsort(choices.v, choices.nmemb, sizeof(struct choice), sort_choices);
> > +}
> > +
> > +static int
> > +sort_choices(const void *p1, const void *p2)
> 
> What do you think about naming this something like `compare_choices` or `cmp_choices` to denote that it serves as a comparison function?
> 
> ---
> https://github.com/thoughtbot/pick/pull/82/files#r48715505

-- 
:wq

mptre:
Great, I will fix the remaining TODO:s in a separate pull request

On Mon, Jan 04, 2016 at 01:46:02AM -0800, Calle Erlandsson wrote:
> Hey @mptre, this is awesome! Thanks for your great work! I left some comments above but other than that this looks great and I'd be happy to merge it. I also agree with your `TODO` notes.
> 
> ---
> https://github.com/thoughtbot/pick/pull/82#issuecomment-168625731

-- 
:wq

src/main.c
calleerlandsson: Thanks! I can't come up with a good name for it. Is `v` for vector? Let's leave it be for now.

calleerlandsson:
Merged as d7d6992. Thanks again, @mptre!

src/main.c
calleerlandsson: I just realized I merged without resolving this one. Fixed in fdaefda.

PR #83, opened 2016-01-04
Simplify `min_match_length`

calleerlandsson:
Simplify `min_match_length` by replacing the existing iterative
algorithm with a recursive one.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -4,6 +4,7 @@
 
 #include <ctype.h>
 #include <err.h>
+#include <limits.h>
 #include <locale.h>
 #include <signal.h>
 #include <stdio.h>
@@ -56,6 +57,7 @@ static struct choice	*selected_choice(void);
 static void		 filter_choices(void);
 static float		 score(char *);
 static size_t		 min_match_length(char *);
+static char		*strcasechr(const char *, char);
 static void		 init_tty(void);
 static int		 tty_putc(int);
 static void		 handle_sigint(int);
@@ -458,7 +460,7 @@ score(char *string)
 	if (string_length == 0)
 		return 0;
 
-	if ((match_length = min_match_length(string)) == 0)
+	if ((match_length = min_match_length(string)) == INT_MAX)
 		return 0;
 
 	return (float)query_length / (float)match_length / (float)string_length;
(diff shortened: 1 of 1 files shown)


calleerlandsson:
@mike-burns, @mptre I would love your feedback on this.

mptre:
I like the idea, however the proposed implementation does not fully utilize the recursive aspect since the computation of the `match_length` function is better off solved using iteration. Instead I would remove the `match_length` auxiliary function and let `min_match_length` do the minimization. Here's a rough outline of my proposal:

```c
static size_t
min_match_length(char *string)
{
	char *start, *end;
	size_t n, m;
	int i;

	start = strchr(string, query[0]);
	if (start == NULL)
		return INT_MAX;

	for (i = 1, end = start + 1; *end && query[i]; end++)
		if (*end == query[i])
			i++;
	n = !query[i] ? end - start : INT_MAX;
	m = min_match_length(string + 1);

	return n < m ? n : m;
}
```

`0` can't be used as a indicator of a missing match since we want to minimize the length therefore `INT_MAX` is used instead. The condition in `score` needs to be adjusted as well:

```c
if ((match_length = min_match_length(string)) == INT_MAX)
```

calleerlandsson:
@mptre That's a great proposal: Makes it even simpler! I have pushed a commit with your proposed algorithm.

> the proposed implementation does not fully utilize the recursive aspect since the computation of the match_length function is better off solved using iteration

Because I'm curious and would love to learn, would you mind elaborating on this? Why is this better off solved using iteration?

mptre:
> Because I'm curious and would love to learn, would you mind elaborating on
> this? Why is this better off solved using iteration?

I can at least try: the prior implementation could easily be solved with a
for-loop in conjunction with a nested for-loop. The overhead of the recursive
call is therefore not justified in my opinion. My proposal could also be solved
with a for-loop but I find using recursion for (min|max)imization problems very
elegant and intuitive since the right answer will "bubble" up in the call stack.

On Tue, Jan 05, 2016 at 01:55:31AM -0800, Calle Erlandsson wrote:
> @mptre That's a great proposal: Makes it even simpler! I have pushed a commit with your proposed algorithm.
> 
> > the proposed implementation does not fully utilize the recursive aspect since the computation of the match_length function is better off solved using iteration
> 
> Because I'm curious and would love to learn, would you mind elaborating on this? Why is this better off solved using iteration?
> 
> ---
> https://github.com/thoughtbot/pick/pull/83#issuecomment-168955132

-- 
:wq

src/main.c
```diff
@@ -479,38 +481,33 @@ score(char *string)
...
-			if (match_length == 0 || match_length > match_position -
-			    match_start + 1)
-				match_length = match_position - match_start + 1;
-		}
-	}
+	char	*start, *end;
+	int	 i;
+	size_t	 length, min;
+
+	if (!(start = strcasechr(string, query[0])))
```
mptre: Why not use `strncasecmp`?

```c
if (!(start = strncasecmp(string, query, 1)))
```

src/main.c
```diff
@@ -479,38 +481,33 @@ score(char *string)
...
+		return INT_MAX;
+	for (i = 1, end = start + 1; *end && query[i]; end++)
+		if (tolower((unsigned char)*end) == tolower((unsigned
```
mptre:
```c
if (strncasecmp(end, &query[i], 1) == 0)
```

src/main.c
calleerlandsson: In this case, `start` has to be a `char *` to the start of the match but `strncasecmp` returns an `int`. I have updated `strcasechr` to use `strncasecmp`, though in af25c5f.

src/main.c
calleerlandsson: Good catch! I didn't think about that. Fixed in af25c5f.

src/main.c
mptre: Ah, my bad.

On Thu, Jan 07, 2016 at 02:14:32AM -0800, Calle Erlandsson wrote:
> > -					if (tolower((unsigned char)string[match_position]) == query_char) {
> > -						++match_position;
> > -						break;
> > -					}
> > -				}
> > -			}
> > -			if (match_length == 0 || match_length > match_position -
> > -			    match_start + 1)
> > -				match_length = match_position - match_start + 1;
> > -		}
> > -	}
> > +	char	*start, *end;
> > +	int	 i;
> > +	size_t	 length, min;
> > +
> > +	if (!(start = strcasechr(string, query[0])))
> 
> In this case, `start` has to be a `char *` to the start of the match but `strncasecmp` returns an `int`. I have updated `strcasechr` to use `strncasecmp`, though in af25c5f.
> 
> ---
> https://github.com/thoughtbot/pick/pull/83/files#r49058819

-- 
:wq

calleerlandsson:
Thanks for the review, @mptre. I'll merge squash and merge this.

PR #85, opened 2016-01-04
Fix Ctrl-K

calleerlandsson:
Pressing Ctrl-K should (according to the man page) delete until the end
of the line. Previously, it would not delete the character under the
cursor and when pressed twice it would bring back the deleted characters
or segfault, depending on the system, due to a memory access error.

This commit fixes the memory access bug and brings the behavior in line
with Bash and Readline which both deletes the character under the
cursor.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -344,7 +344,7 @@ selected_choice(void)
 			delete_between(
 			    query,
 			    query_length,
-			    cursor_position + 1,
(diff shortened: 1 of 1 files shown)


mike-burns:
Simple fix. LGTM.

calleerlandsson:
Thanks, @mike-burns. I'll merge this.

PR #86, opened 2016-01-04
Scroll query horizontally as needed

calleerlandsson:
Previously when a query that wouldn't fit within the number of columns
of the tty was entered, the query input field would be unable to show
the full query.

This commit adds horizontal scrolling to the query input field. When the
cursor is moved outside of the currently visible part of the query, the
query is scrolled to show the previously invisible part.


mike-burns:
I now understand what is on our office whiteboard.

The code looks good but I only gave it a quick glance without trying it.

calleerlandsson:
Thanks, @mike-burns. I've rebase this on the current master and it works great on my system. I'll merge this.

PR #87, opened 2016-01-04
Refactor and reformat for clarity

calleerlandsson:
- Inline one line functions which names are not needed for clarification
  of their functionality.
- Simplify initialization in `selected_choice`.


mike-burns:
Agreed: just as easy to read but with less code. LGTM.

calleerlandsson:
Thanks, @mike-burns! I'll merge this.

PR #88, opened 2016-01-05
Remove SLIST

mptre:
Since d7d6992 this is no longer needed.


calleerlandsson:
I like these diff stats: `+0 -662`

Great work, @mptre! Merged as 6c3e49a685fe013b4cc25fea16e08ea79be47fa0.

PR #89, opened 2016-01-05
Remove _WITH_GETLINE definition

mptre:
Regression from bf58c61.


calleerlandsson:
Good catch, @mptre! Merged as 52e1b8e02b710e040d8ca3e32fc5d41cb6716f85.

PR #92, opened 2016-01-07
Remove static from function definitions

mptre:
If the function is declared as static there's no need to repeat the
declaration.


calleerlandsson:
Today I learned! Thanks, @mptre. Merged as 5d2effe.

PR #93, opened 2016-01-07
Rename main.c -> pick.c

mptre:
Since this makes more sense.


calleerlandsson:
I don't have strong opinion about this but am fine with the rename. Merged as 853f029.

PR #94, opened 2016-01-08
Reduce the number of scoring states

mptre:
Assigning a score of 1 to all choices is used to force output of all choices when the query is empty. Instead modify the loop condition to check if the query is empty and then bypass the score condition. A pleasant side-effect of this change is that each score is either 0 or >0 which implies presence of a score.


calleerlandsson:
Hey, @mptre, great work! I've merged this as 6a3a838, 3fdb269, 931f64c.

PR #95, opened 2016-01-15
Fix off-by-one bug

mptre:
My assumption was wrong in this case.

Can be verified:
```bash
printf 'a\nb\n' | LINES=2 pick
```

Prior this change an empty lines was emitted.


calleerlandsson:
Good catch!

PR #97, opened 2016-01-25
Highlight the matched substring

mptre:
By using the underline capability the matched substring for all choices
is highlighted. This change gives a better notion of the actual match
and allows one to easily spot matched outliers.

The printing of choices is rewritten which removes the need to allocate
a new string for every matched choice and usage of two compat functions.


calleerlandsson:
@mptre Very cool! Those are amazing diff stats for an added feature. Works great on my machine. Let me take a closer look at the code on Friday.

calleerlandsson:
@mptre Sorry for taking so long. I've looked over the code and it looks great! I'll merge this.

mptre:
Great! Time for a proper release?

calleerlandsson:
Yes, definitely! I'll use the current `master` for the coming week and then release next Friday.

I'm leaning on 1.3 for the version number. What are your thoughts?

mptre:
On Fri, Feb 05, 2016 at 07:32:07AM -0800, Calle Erlandsson wrote:
> Yes, definitely! I'll use the current `master` for the coming week and then release next Friday.
> 
> I'm leaning on 1.3 for the version number. What are your thoughts?

None of the changes is breaking therefore 1.3 makes sense.

PR #101, opened 2016-02-18
Include <sys/ttydefaults.h> for the CTRL macro

fornwall:
With acd349122e36a49e17e3c04a54cfb08bea9e50fd the CTRL macro started
being used but <sys/ttydefaults.h> was never explicitly included, which
broke the build when using the Android NDK or the musl libc.

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -2,6 +2,7 @@
 #include "config.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


calleerlandsson:
Hey, @fornwall, thanks for catching and fixing this! I'll merge this.

PR #102, opened 2016-02-19
Only count printable characters

### LINKED ISSUE #100: Colors in output disrupt line alignment
I had a script working in v1.2.1 that is useful for selecting from recently checked out git branches. I was using color to divide some of the columns. I've narrowed the issue down to being color specifically. This occurs in both bash and zsh for me. Here's an example that produces odd output:

```bash
RED='\033[0;31m'
NC='\033[0m'
printf "${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n" | pick
```

So, you can see we've created a list of colorized items to pipe to pick. Here is what the output looks like on my system for some reason. The issue first came up after installing v1.3.

image

Thank you guys so very much for this great tool, I've applied it to all sorts of things, and makes my work merrier several times a day. You guys rock!

calleerlandsson:
When counting how many characters are left to print from the end of a
choice to the right-most column of the tty, disregard escape sequences.

This fixes #100.

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -38,6 +38,7 @@
 #define EX_SIG 128
 #define EX_SIGINT (EX_SIG + SIGINT)
 
+#define MAX(x, y) ((x) > (y) ? (x) : (y))
 #define MIN(x, y) ((x) < (y) ? (x) : (y))
 #define tty_putp(capability) do { \
 	if (tputs(capability, 1, tty_putc) == ERR) \
@@ -48,6 +49,7 @@ struct choice {
 	char	*description;
 	char	*string;
 	size_t	 length;
+	size_t	 printable_length;
 	size_t	 match_start;
 	size_t	 match_end;
 	float	 score;
@@ -57,6 +59,7 @@ __dead static void		 usage(void);
 __dead static void		 version(void);
 static void			 get_choices(void);
 static char			*eager_strpbrk(const char *, const char *);
+static size_t			 count_printable(char *, size_t length);
 static void			 put_choice(const struct choice *);
 static const struct choice	*selected_choice(void);
 static void			 filter_choices(void);
@@ -218,6 +221,8 @@ get_choices(void)
 		choices.v[choices.length].length = stop - start;
 		choices.v[choices.length].string = start;
 		choices.v[choices.length].description = description;
+		choices.v[choices.length].printable_length =
+		    count_printable(start, stop - start);
 		choices.v[choices.length].match_start = 0;
 		choices.v[choices.length].match_end = 0;
 		choices.v[choices.length].score = 0;
@@ -244,6 +249,25 @@ eager_strpbrk(const char *string, const char *separators)
 		ptr = tmp_ptr++;
 
 	return ptr;
+}
+
+size_t
+count_printable(char *str, size_t length)
+{
+	int	in_esc_seq = 0, printable = 0;
+	size_t	i;
+
+	for (i = 0; i < length; i++) {
+		if (in_esc_seq) {
+			in_esc_seq = (str[i - 1] == ESCAPE && str[i] == '[') ||
+			    str[i] < '@' || str[i] > '~';
+		} else {
+			in_esc_seq = str[i] == ESCAPE;
+			if (!in_esc_seq)
+				printable++;
+		}
+	}
+	return printable;
 }
 
 void
@@ -597,7 +621,7 @@ print_query(char *query, size_t length, size_t position, size_t scroll)
 int
 print_choices(int selection)
 {
-	int		 i, j;
+	int		 i, j, k;
 	size_t		 length, query_length;
 	struct choice	*choice;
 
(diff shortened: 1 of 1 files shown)


calleerlandsson:
```
~/Projects/pick$ RED='\033[0;31m'
~/Projects/pick$ NC='\033[39m'
~/Projects/pick$ printf "${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n${RED}pick${NC} is the handiest\n" | src/pick
```

Before:

screen shot 2016-02-19 at 14 43 00

After:

screen shot 2016-02-19 at 14 43 26

src/pick.c
```diff
@@ -37,6 +37,8 @@
 #define EX_SIG 128
 #define EX_SIGINT (EX_SIG + SIGINT)
 
+#define IN_ESC_SEQ(s, i) ((s[i - 1] == ESCAPE && s[i] == '[') || s[i] < '@' || \
```
mike-burns: I worry about expansion -- should `s` and `i` be wrapped in parens?

src/pick.c
```diff
@@ -612,33 +614,54 @@ print_choices(int selection)
...
+		p = 0;
+
 		for (j = 0, length = MIN(choice->match_start, (size_t)columns);
 		     j < (ssize_t)length;
-		     j++)
+		     j++) {
+			if (esc)
+				esc = IN_ESC_SEQ(choice->string, j);
+			else if (!(esc = (choice->string[j] == ESCAPE)))
+				p++;
```
mike-burns: Maybe these could be extracted like:

```c
p = skip_esc(p, choice->string, j, &esc);
```

So I guess:

```c
int
skip_esc(int p, char *str, int j, int *esc)
{
  if (*esc) {
    *esc = IN_ESC_SEQ(str, j);
    return p;
  } else if(!(*esc = (str[j] == ESCAPE)))
    return p + 1;
}
```

I don't love that `(!(esc = (choice->string[j] == ESCAPE)))` line because I was skimming for where `esc` is set and instead I have to carefully read this line. How about pulling out the assignment from the comparison?

src/pick.c
calleerlandsson: Good point. Parens added in e1b7bda.

src/pick.c
calleerlandsson: Great idea @mike-burns! What do you think of 8747705?

mike-burns:
Rad. LGTM.

mptre:
Could the number of printable/non-printable characters for each choice
be calculated once in `get_choices`?

On Fri, Feb 19, 2016 at 07:52:01AM -0800, Mike Burns wrote:
> Rad. LGTM.
> 
> ---
> https://github.com/thoughtbot/pick/pull/102#issuecomment-186272413

-- 
:wq

calleerlandsson:
@mptre That's a great idea. I'll look into it on Friday!

calleerlandsson:
I've updated the implementation to only calculate the number of printable characters once. Let me know what you think, @mike-burns, @mptre.

src/pick.c
```diff
@@ -48,6 +49,7 @@ struct choice {
 	char	*description;
 	char	*string;
 	size_t	 length;
+	size_t	 printable;
```
mptre: Could it be renamed to `printable_length` for consistency and add a comment on the definition of a printable character?

src/pick.c
```diff
@@ -246,6 +251,25 @@ eager_strpbrk(const char *string, const char *separators)
...
 }
 
+size_t
+count_printable(char *str, size_t length)
+{
+	int	in_esc_seq = 0, printable = 0;
+	size_t	i;
+
+	for (i = 0; i < length; i++) {
+		if (in_esc_seq)
```
mptre: I would wrap the if-statement in braces.

mptre:
Other than my comments it looks good, great work.

On Fri, Feb 26, 2016 at 01:48:45AM -0800, Calle Erlandsson wrote:
> I've updated the implementation to only calculate the number of printable characters once. Let me know what you think, @mike-burns, @mptre.
> 
> ---
> https://github.com/thoughtbot/pick/pull/102#issuecomment-189195430

-- 
:wq

calleerlandsson:
Thanks! I'll resolve your comments and merge this on Friday.

src/pick.c
calleerlandsson: Yeah, that's a great idea! Done in 3c2bec4.

src/pick.c
calleerlandsson: Good call! Done in 50f4b06.

calleerlandsson:
Thanks for reviewing @mike-burns and @mptre. I'll squash and merge this.

PR #106, opened 2016-03-10
Add Unicode support

mptre:
Work in progress, would really appreciate some help with testing. A few notes in no particular order:

- The `get_key` function recognizes both `\033[` and `\033O`, which terminal omits the later?

- I changed `BACKSPACE` from `263` to `8` which is the decimal number for the backspace character. Is this a bad idea? Where did `263` come from?

- Comparing Unicode characters is not case insensitive. Using one of the wide character comparision functions seems to best way to do it at glance since it's not as trivial as comparing ASCII characters. Input on this would be appreciated.

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -14,6 +14,7 @@
 #include <sysexits.h>
 #include <termios.h>
 #include <unistd.h>
+#include <wchar.h>
 
 #ifdef HAVE_NCURSESW_H
 #include <ncursesw/curses.h>
@@ -28,7 +29,7 @@
 #define DEL 127
 #define ENTER 10
 #define ALT_ENTER 266
-#define BACKSPACE 263
+#define BACKSPACE 8
 #define UP 259
 #define DOWN 258
 #define RIGHT 261
@@ -39,7 +40,6 @@
 #define EX_SIGINT (EX_SIG + SIGINT)
 
 #define MAX(x, y) ((x) > (y) ? (x) : (y))
-#define MIN(x, y) ((x) < (y) ? (x) : (y))
 #define tty_putp(capability) do { \
 	if (tputs(capability, 1, tty_putc) == ERR) \
 		errx(1, #capability ": unknown terminfo capability"); \
@@ -65,19 +65,21 @@ static const struct choice	*selected_choice(void);
 static void			 filter_choices(void);
 static void			 score(struct choice *);
 static int			 min_match(const char *, size_t, size_t *, size_t *);
-static char			*strcasechr(const char *, char);
+static char			*strcasechr(const char *, const char *);
 static void			 init_tty(void);
 static int			 tty_putc(int);
 static void			 handle_sigint(int);
 static void			 restore_tty(void);
 static void			 put_line(char *, int);
 static int			 print_choices(int);
-static int			 get_key(void);
+static size_t			 get_key(char *, size_t);
 static int			 tty_getc(void);
 static void			 delete_between(char *, size_t, size_t, size_t);
 static void			 free_choices(void);
 static void			 print_query(char *, size_t, size_t, size_t);
 static int			 choicecmp(const void *, const void *);
+static int			 isu8cont(unsigned char);
+static int			 isu8start(unsigned char);
 
 static FILE		*tty_in;
 static FILE		*tty_out;
@@ -105,6 +107,8 @@ main(int argc, char **argv)
 	int	option;
 
 	use_alternate_screen = getenv("VIM") == NULL;
+
+	setlocale(LC_CTYPE, "");
 
 	while ((option = getopt(argc, argv, "dhoq:SvxX")) != -1) {
 		switch (option) {
@@ -287,9 +291,10 @@ put_choice(const struct choice *choice)
 const struct choice *
 selected_choice(void)
 {
-	int		 key, selection = 0, visible_choices_count;
+	char		 key[6];
+	int		 selection = 0, visible_choices_count;
 	int		 word_position;
-	size_t		 cursor_position, query_length, scroll;
+	size_t		 cursor_position, length, query_length, scroll;
 
 	cursor_position = query_length = strlen(query);
 
@@ -307,9 +312,10 @@ selected_choice(void)
 
 	for (;;) {
 		fflush(tty_out);
-		key = get_key();
+		memset(key, 0, sizeof(key));
+		length = get_key(key, sizeof(key));
 
-		switch (key) {
+		switch (*((int *)key)) {
 		case ENTER:
 			if (visible_choices_count > 0) {
 				restore_tty();
@@ -328,13 +334,16 @@ selected_choice(void)
 		case BACKSPACE:
 		case DEL:
 			if (cursor_position > 0) {
+				for (length = 1;
+				    isu8cont(query[cursor_position - length]);
+				    length++);
 				delete_between(
 				    query,
 				    query_length,
-				    cursor_position - 1,
+				    cursor_position - length,
 				    cursor_position);
-				--cursor_position;
-				--query_length;
+				cursor_position -= length;
+				query_length -= length;
 				filter_choices();
 				selection = 0;
 			}
@@ -342,12 +351,15 @@ selected_choice(void)
 			break;
 		case CTRL('D'):
 			if (cursor_position < query_length) {
+				for (length = 1;
+				    isu8cont(query[cursor_position + length]);
+				    length++);
 				delete_between(
 				    query,
 				    query_length,
 				    cursor_position,
-				    cursor_position + 1);
-				--query_length;
+				    cursor_position + length);
+				query_length -= length;
 				filter_choices();
 				selection = 0;
 			}
@@ -375,25 +387,26 @@ selected_choice(void)
 			selection = 0;
 			break;
 		case CTRL('W'):
-			if (cursor_position > 0) {
-				for (word_position = cursor_position - 1;
-				    word_position > 0;
-				    --word_position) {
-					if (query[word_position] != ' ' &&
-					    query[word_position - 1] == ' ')
-						break;
-				}
+			if (cursor_position == 0)
+				break;
 
-				delete_between(
-				    query,
-				    query_length,
-				    word_position,
-				    cursor_position);
-				query_length -= cursor_position - word_position;
-				cursor_position = word_position;
-				filter_choices();
-				selection = 0;
+			for (word_position = cursor_position;;) {
+				while (isu8cont(query[--word_position]));
+				if (word_position < 1)
+					break;
+				if (query[word_position] != ' '
+				    && query[word_position - 1] == ' ')
+					break;
 			}
+			delete_between(
+			    query,
+			    query_length,
+			    word_position,
+			    cursor_position);
+			query_length -= cursor_position - word_position;
+			cursor_position = word_position;
+			filter_choices();
+			selection = 0;
 			break;
 		case CTRL('A'):
 			cursor_position = 0;
@@ -413,41 +426,41 @@ selected_choice(void)
 			break;
 		case CTRL('B'):
 		case LEFT:
-			if (cursor_position > 0)
-				--cursor_position;
+			while (cursor_position > 0
+			    && isu8cont(query[--cursor_position]));
 			break;
 		case CTRL('F'):
 		case RIGHT:
-			if (cursor_position < query_length)
-				++cursor_position;
+			while (cursor_position < query_length
+			    && isu8cont(query[++cursor_position]));
 			break;
 		default:
-			if (key > 31 && key < 127) { /* Printable chars */
-				if (cursor_position < query_length) {
-					memmove(
-					    query + cursor_position + 1,
-					    query + cursor_position,
-					    query_length - cursor_position);
-				}
+			if (!isu8start(key[0]) && !isprint(key[0]))
+				continue;
 
-				query[cursor_position++] = key;
-				query[++query_length] = '\0';
-				filter_choices();
-				selection = 0;
+			if (query_size < query_length + length) {
+				query_size = 2*query_length + length;
+				if ((query = reallocarray(query, query_size,
+					    sizeof(char))) == NULL)
+					err(1, NULL);
 			}
+
+			if (cursor_position < query_length)
+				memmove(query + cursor_position + length,
+					query + cursor_position,
+					query_length - cursor_position);
+
+			memcpy(query + cursor_position, key, length);
+			cursor_position += length;
+			query_length += length;
+			query[query_length] = '\0';
+			filter_choices();
+			selection = 0;
 
 			break;
 		}
 
 		tty_putp(cursor_invisible);
-
-		if (query_length == query_size - 1) {
-			query_size += query_size;
-
-			query = reallocarray(query, query_size, sizeof(*query));
-			if (query == NULL)
-				err(1, "reallocarray");
-		}
 
 		visible_choices_count = print_choices(selection);
 		tty_putp(clr_eos);
@@ -508,16 +521,19 @@ score(struct choice *choice)
 int
 min_match(const char *string, size_t offset, size_t *start, size_t *end)
 {
-	char	*s, *e;
-	int	 i;
+	char	 *s, *e, *q;
 
-	if (!(s = strcasechr(string + offset, query[0])))
+	q = query;
+	if ((s = e = strcasechr(&string[offset], q)) == NULL)
 		return 0;
-	for (i = 1, e = s + 1; *e && query[i]; e++)
-		if (strncasecmp(e, query + i, 1) == 0)
-			i++;
-	if (query[i])
-		return 0;
+
+	for (;;) {
+		for (e++, q++; isu8cont(*q); e++, q++);
+		if (*q == '\0')
+			break;
+		if ((e = strcasechr(e, q)) == NULL)
+			return 0;
+	}
 
 	*start = s - string;
 	*end = e - string;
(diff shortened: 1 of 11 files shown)


calleerlandsson:
Looks really cool, @mptre! I'll look closer at the code and try it out as soon as I get some time.

I have no idea where `263` came from originally but expect it was from trying it out and printing whatever input I got.

src/pick.c
```diff
@@ -409,41 +415,43 @@ selected_choice(void)
...
-				query[cursor_position++] = key;
-				query[++query_length] = '\0';
-				filter_choices();
-				selection = 0;
+			if (query_size < query_length + length) {
+				do
+					query_size *= 2;
+				while (query_size < query_length + length);
+				query = reallocarray(
+				    query, query_size, sizeof(char));
```
mike-burns: OpenBSD's reallocarray(3) says:

> Do not adjust the variable describing how much memory has been allocated until the allocation has been successful. This can cause aberrant program behavior if the incorrect size value is used. In most cases, the above sample will also result in a leak of memory. As stated earlier, a return value of NULL indicates that the old object still remains allocated.

http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man3/calloc.3

They show this with `realloc` but then later with `reallocarray`.

src/pick.c
```diff
@@ -409,41 +415,43 @@ selected_choice(void)
...
+				if (query == NULL)
+					err(1, "reallocarray");
```
mike-burns: Passing `NULL` as the second argument to err(3) will print a default message, which (should, hopefully) include the text "reallocarray".

mptre:
On Thu, Mar 10, 2016 at 02:38:19AM -0800, Calle Erlandsson wrote:
> Looks really cool, @mptre! I'll look closer at the code and try it out as soon as I get some time.
> 
> I have no idea where `263` came from originally but expect it was from trying it out and printing whatever input I got.

I see, the change is then reasonable and can be kept?.

src/pick.c
mptre: On Thu, Mar 10, 2016 at 02:42:33AM -0800, Mike Burns wrote:
> > -					memmove(
> > -					    query + cursor_position + 1,
> > -					    query + cursor_position,
> > -					    query_length - cursor_position);
> > -				}
> > -
> > -				query[cursor_position++] = key;
> > -				query[++query_length] = '\0';
> > -				filter_choices();
> > -				selection = 0;
> > +			if (query_size < query_length + length) {
> > +				do
> > +					query_size *= 2;
> > +				while (query_size < query_length + length);
> > +				query = reallocarray(
> > +				    query, query_size, sizeof(char));
> 
> OpenBSD's reallocarray(3) says:
> 
> > Do not adjust the variable describing how much memory has been allocated until the allocation has been successful. This can cause aberrant program behavior if the incorrect size value is used. In most cases, the above sample will also result in a leak of memory. As stated earlier, a return value of NULL indicates that the old object still remains allocated.
> 
> http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man3/calloc.3
> 
> They show this with `realloc` but then later with `reallocarray`.

I'm aware of this. Since it terminates on `reallocarray` failure: is it
really necessary?

src/pick.c
mptre: On Thu, Mar 10, 2016 at 02:43:18AM -0800, Mike Burns wrote:
> > -					    query + cursor_position,
> > -					    query_length - cursor_position);
> > -				}
> > -
> > -				query[cursor_position++] = key;
> > -				query[++query_length] = '\0';
> > -				filter_choices();
> > -				selection = 0;
> > +			if (query_size < query_length + length) {
> > +				do
> > +					query_size *= 2;
> > +				while (query_size < query_length + length);
> > +				query = reallocarray(
> > +				    query, query_size, sizeof(char));
> > +				if (query == NULL)
> > +					err(1, "reallocarray");
> 
> Passing `NULL` as the second argument to err(3) will print a default message, which (should, hopefully) include the text "reallocarray".

Correct, thanks. Fixed in b8961ce.

src/pick.c
mptre: A somewhat related question: do you use any program or Vim options to
ensure KNF (style(9)) compliance?

mike-burns: In vim:

```
set cinoptions=:0,t0,+4,(4
au BufNewFile,BufReadPost *.{c,h,cc,cpp,hpp,y,l} setl noet sw=0 ts=8
```

mptre: On Thu, Mar 10, 2016 at 04:56:17AM -0800, Mike Burns wrote:
> In vim:
> 
> ```
> set cinoptions=:0,t0,+4,(4
> au BufNewFile,BufReadPost *.{c,h,cc,cpp,hpp,y,l} setl noet sw=0 ts=8
> ```

Ty!

calleerlandsson:
Trying this out on OS X, it seems to work really well! I was happily surprised even this works:

screen shot 2016-03-21 at 17 07 20

Like we talked about the last time we met, the case insensitive comparison doesn't work for non-ASCII characters. Even though this doesn't break the current behavior, I do think users will expect it.

I haven't had time to dive into how wchars work yet and won't probably have time during the week. I guess I know what I'll be doing for easter! :chicken:

Thanks again for spending time on this @mptre!

mptre:
I've been digging through the OpenBSD source for uses of wchar in order
to resolve how-to use it in pick. Will hopefully have something working
later this week.

BTW: there's a test case for emojis :)

calleerlandsson:
Cool!

Ha! Yeah, I saw that one.

mptre:
Just pushed a commit which fixes case insensitive comparison even for Unicode character. The test suite made the migration to wchar a pleasant experience. However I'm going to run this version locally for a couple of days in order to ensure no regression was introduced.

As for the commits I want to do a proper squash before merging.

There is still one unanswered question: the get_key function recognizes both `\033[` and `\033O`, which terminal omits the later?

calleerlandsson:
Awesome! Thanks @mptre!

Could that be when running inside of Vim using the `system` function? I remember codes sent by the arrow keys changed inside Vim.

When trying this out locally i observed this behavior:

```
touch ö
ls | src/pick # ö is part of the list
type ö # no results
backspace and type o # ö are in the results
exit pick
printf "ö\n" | src/pick
type ö # ö are in the results
```

Could this be because of how my filesystem works?

mptre:
On Tue, Mar 29, 2016 at 08:21:01AM -0700, Calle Erlandsson wrote:
> Awesome! Thanks @mptre!
> 
> Could that be when running inside of Vim using the `system` function? I remember codes sent by the arrow keys changed inside Vim.

Thanks, verified.

> 
> When trying this out locally i observed this behavior:
> 
> ```
> touch ö
> ls | src/pick # ö is part of the list
> type ö # no results
> backspace and type o # ö are in the results
> exit pick
> printf "ö\n" | src/pick
> type ö # ö are in the results
> ```
> 
> Could this be because of how my filesystem works?

Are you using OS X? Could you provide the output of the following
command:

  mkdir tmp
  cd tmp
  touch ö
  ( \ls ; printf '\x00ö' ) | hexdump -C

calleerlandsson:
@mptre Yes I am.

Here's the output:

```
~$ mkdir tmp
~$ cd tmp
~/tmp$ touch ö
~/tmp$ ( \ls ; printf '\x00ö' ) | hexdump -C
00000000  6f cc 88 0a 00 c3 b6                              |o......|
00000007
~/tmp$ 
```

It seems like they produce different bytes. Reading <https://en.wikipedia.org/wiki/HFS_Plus>, there definitely seems to be some oddities to how filenames are encoded. If this is a HFS+ specific problem I see no reason for pick accounting for it.

mptre:
On Wed, Mar 30, 2016 at 12:40:28AM -0700, Calle Erlandsson wrote:
> @mptre Yes I am.
> 
> Here's the output:
> 
> ```
> ~$ mkdir tmp
> ~$ cd tmp
> ~/tmp$ touch ö
> ~/tmp$ ( \ls ; printf '\x00ö' ) | hexdump -C
> 00000000  6f cc 88 0a 00 c3 b6                              |o......|
> 00000007
> ~/tmp$ 
> ```
> 
> It seems like they produce different bytes. Reading <https://en.wikipedia.org/wiki/HFS_Plus>, there definitely seems to be some oddities to how filenames are encoded. If this is a HFS+ specific problem I see no reason for pick accounting for it.

Thanks for the report. It seems like HFS stores Unicode character in
decomposed form as opposed of precomposed[1]. I'm not sure if any
wchar-related function can normalize characters.

[1] http://stackoverflow.com/a/6153713

src/pick.c
```diff
@@ -65,19 +65,21 @@ static const struct choice	*selected_choice(void);
 static void			 filter_choices(void);
 static void			 score(struct choice *);
 static int			 min_match(const char *, size_t, size_t *, size_t *);
-static char			*strcasechr(const char *, char);
+static char			 *strcasechr(const char *, const char *);
```
calleerlandsson: Should we remove the space in front of the first `*` on this line?

In the past, I've tried to align the declared function names for readability. The added space in front of the functions not returning a pointer is to account for the asterisk for those that do. This is described in OpenBSD's style(9):

> Prototypes may have an extra space after a tab to enable function names to line up:
> ```
> static char	*function(int, const char *); 
> static void	 usage(void);
> ```

src/pick.c
```diff
@@ -729,3 +792,15 @@ free_choices(void)
 	free(choices.v);
 	free(input.string);
 }
+
+int
+isu8cont(unsigned char c)
```
calleerlandsson: Does this function return whether or not the byte is the continuation of an utf-8 character?

src/pick.c
```diff
@@ -729,3 +792,15 @@ free_choices(void)
...
+{
+	return (c & (0x80 | 0x40)) == 0x80;
+}
+
+int
+isu8start(unsigned char c)
```
calleerlandsson: Does this function return whether or not the byte is the continuation of an utf-8 character?

src/pick.c
```diff
@@ -608,62 +643,70 @@ put_line(char *string, int length)
...
 			 * terminating null character if descriptions is
 			 * enabled. */
-			if (tty_putc(choice->string[j] == '\0'
-				     ? ' ' : choice->string[j]) == EOF)
-				err(1, "tty_putc");
+			if (choice->string[j] == '\0') {
+				if (tty_putc(' ') == EOF)
+					err(1, "tty_putc");
+			} else {
+				if (tty_putc(choice->string[j]) == EOF)
```
calleerlandsson: Could this `if` be joined with the above `else` to reduce indentation?

calleerlandsson:
I've read through the code and I think it looks great! I've added some comments above.

In general it seems like this adds quite a bit of complexity but I also think that is due to the complexity of UTF-8. I wonder if adding some comments about the code that relates to how UTF-8 works would help or if readers will just have to dig into UTF-8 documentation.

One thing that we seem to do a lot is to loop through the bytes that makes up a single character. Maybe this could be extracted into a well-named macro?

When running this locally I found the following things. I haven't been able to come up with fixes for them so if you have any advice on how to approach them I'd really appreciate it!

- My delete key no longer works. Instead of deleting the character under the cursor it moves the cursor forward by 4 spaces. How do I easiest find out what byte sequence my terminal sends when I press the delete key?

- Some non-printable chars seems to be recognized as printable chars. For example, when pressing page up and page down the cursor is moved forward by 4 spaces. Instead I would expect nothing to happen. Previously we checked if the character was in a printable range before adding it to the query and I understand it is not that simple anymore. What can we do to improve this?

- When highlighting the last choice in the list, the last column of the terminal isn't highlighted. I think this might be an off by one error somewhere but I haven't found it:
screen shot 2016-04-08 at 11 06 07
screen shot 2016-04-08 at 11 06 25

src/pick.c
mptre: On Fri, Apr 08, 2016 at 01:24:34AM -0700, Calle Erlandsson wrote:
> > @@ -65,19 +65,21 @@ static const struct choice	*selected_choice(void);
> >  static void			 filter_choices(void);
> >  static void			 score(struct choice *);
> >  static int			 min_match(const char *, size_t, size_t *, size_t *);
> > -static char			*strcasechr(const char *, char);
> > +static char			 *strcasechr(const char *, const char *);
> 
> Should we remove the space in front of the first `*` on this line?
> 
> In the past, I've tried to align the declared function names for readability. The added space in front of the functions not returning a pointer is to account for the asterisk for those that do. This is described in OpenBSD's style(9):
> 
> > Prototypes may have an extra space after a tab to enable function names to line up:
> > ```
> > static char	*function(int, const char *); 
> > static void	 usage(void);
> > ```

Nice catch, fixed in 0bd3350.

src/pick.c
mptre: On Fri, Apr 08, 2016 at 01:41:19AM -0700, Calle Erlandsson wrote:
> > @@ -729,3 +792,15 @@ free_choices(void)
> >  	free(choices.v);
> >  	free(input.string);
> >  }
> > +
> > +int
> > +isu8cont(unsigned char c)
> 
> Does this function return whether or not the byte is the continuation of an utf-8 character?

It returns 1 if the given byte is a UTF-8 continuation byte.

src/pick.c
mptre: On Fri, Apr 08, 2016 at 01:41:23AM -0700, Calle Erlandsson wrote:
> > @@ -729,3 +792,15 @@ free_choices(void)
> >  	free(choices.v);
> >  	free(input.string);
> >  }
> > +
> > +int
> > +isu8cont(unsigned char c)
> > +{
> > +	return (c & (0x80 | 0x40)) == 0x80;
> > +}
> > +
> > +int
> > +isu8start(unsigned char c)
> 
> Does this function return whether or not the byte is the continuation of an utf-8 character?

It returns 1 if the given byte marks the start of a UTF-8 character: the
two MSBs are both set to 1 as opposed of a continuation byte where the
two MSBs are equal to `10'.

src/pick.c
mptre: On Fri, Apr 08, 2016 at 01:43:14AM -0700, Calle Erlandsson wrote:
> >  			/* A null character will be present before the
> >  			 * terminating null character if descriptions is
> >  			 * enabled. */
> > -			if (tty_putc(choice->string[j] == '\0'
> > -				     ? ' ' : choice->string[j]) == EOF)
> > -				err(1, "tty_putc");
> > +			if (choice->string[j] == '\0') {
> > +				if (tty_putc(' ') == EOF)
> > +					err(1, "tty_putc");
> > +			} else {
> > +				if (tty_putc(choice->string[j]) == EOF)
> 
> Could this `if` be joined with the above `else` to reduce indentation?

Nice catch, fixed in 727b8f6.

mptre:
On Fri, Apr 08, 2016 at 02:07:13AM -0700, Calle Erlandsson wrote:
> I've read through the code and I think it looks great! I've added some comments above.
> 
> In general it seems like this adds quite a bit of complexity but I also think that is due to the complexity of UTF-8. I wonder if adding some comments about the code that relates to how UTF-8 works would help or if readers will just have to dig into UTF-8 documentation.

I could add comments on the purpose and behavior UTF-8 related functions
if you find that of value.

> One thing that we seem to do a lot is to loop through the bytes that makes up a single character. Maybe this could be extracted into a well-named macro?

Sure, however I haven't been able to generalize the logic. Any ideas are
much appreciated.

> When running this locally I found the following things. I haven't been able to come up with fixes for them so if you have any advice on how to approach them I'd really appreciate it!
> 
> - My delete key no longer works. Instead of deleting the character under the cursor it moves the cursor forward by 4 spaces. How do I easiest find out what byte sequence my terminal sends when I press the delete key?

Did your delete key trigger the same behavior as CTRL-D? Mine does not,
even with the current state of the master branch.

Run `cat' and press your delete key to find the sequence. It is most
probably equal to:

  tput kD | hexdump -C

> - Some non-printable chars seems to be recognized as printable chars. For example, when pressing page up and page down the cursor is moved forward by 4 spaces. Instead I would expect nothing to happen. Previously we checked if the character was in a printable range before adding it to the query and I understand it is not that simple anymore. What can we do to improve this?

Fixed in 37d43b6.

> - When highlighting the last choice in the list, the last column of the terminal isn't highlighted. I think this might be an off by one error somewhere but I haven't found it:

Nice catch. This is broken in master branch as well. I'm going to look
into this soon and use another branch since it isn't related to the
Unicode code.

calleerlandsson:
Hey @mptre, thanks for fixing the non-printable chars issue! Trying it now, it seems to work really good!

> I could add comments on the purpose and behavior UTF-8 related functions if you find that of value.

That would be great!

> Sure, however I haven't been able to generalize the logic. Any ideas are much appreciated.

After trying for a bit, I haven't found a good generalization either. Let's leave it for now.

> Did your delete key trigger the same behavior as CTRL-D? Mine does not, even with the current state of the master branch.

After trying it on master, I realize it never worked this way so let's leave this for now as well.

> Nice catch. This is broken in master branch as well. I'm going to look into this soon and use another branch since it isn't related to the Unicode code.

Sounds great!

I think this branch is ready to merge! Would you like to squash it somewhat and rebase it on the current master before I merge it?

mptre:
On Mon, Apr 18, 2016 at 05:17:50AM -0700, Calle Erlandsson wrote:
> Hey @mptre, thanks for fixing the non-printable chars issue! Trying it now, it seems to work really good!
> 
> > I could add comments on the purpose and behavior UTF-8 related functions if you find that of value.
> 
> That would be great!
> 
> > Sure, however I haven't been able to generalize the logic. Any ideas are much appreciated.
> 
> After trying for a bit, I haven't found a good generalization either. Let's leave it for now.
> 
> > Did your delete key trigger the same behavior as CTRL-D? Mine does not, even with the current state of the master branch.
> 
> After trying it on master, I realize it never worked this way so let's leave this for now as well.
> 
> > Nice catch. This is broken in master branch as well. I'm going to look into this soon and use another branch since it isn't related to the Unicode code.
> 
> Sounds great!
> 
> I think this branch is ready to merge! Would you like to squash it somewhat and rebase it on the current master before I merge it?

Squashed and rebased. I want to take a last look at the changes later
this week before merging.

mptre:
I simplified the query memory reallocation and chomped up a long line. I
think this is ready to be merged by now.

calleerlandsson:
Looks great! Do you want to squash in the fixup commits before I merge or should I?

mptre:
On Thu, Apr 21, 2016 at 02:40:23AM -0700, Calle Erlandsson wrote:
> Looks great! Do you want to squash in the fixup commits before I merge or should I?

Squashed!

calleerlandsson:
Thanks @mptre, I'll merge this.

PR #108, opened 2016-04-08
Add an option for disable sorting

### LINKED ISSUE #107: Stable sort option
The default sort is good for some use cases, but if you're trying to filter things that are in chronological order (like git commits) it becomes a bit lackluster. Having an option for stable sorting would be very valuable for picking things from an externally-ordered list.

calleerlandsson:
When filtering an already sorted list of choices, for example a
chronologically sorted list of git commits, finding a specific one is
easier if pick doesn't sort the choices according to their score.

This commit adds an `-S` optional flag to disable sorting.

This closes #107.

### DIFF
--- a/man/man1/pick.1
+++ b/man/man1/pick.1
@@ -6,7 +6,7 @@
 .Nd fuzzy select anything
 .Sh SYNOPSIS
 .Nm
-.Op Fl hv
+.Op Fl hvS
 .Op Fl d Op Fl o
 .Op Fl x | Fl X
 .Op Fl q Ar query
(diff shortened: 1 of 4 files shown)


man/man1/pick.1
```diff
@@ -26,6 +26,10 @@ supports these options and arguments:
 output a help message and exit.
 .It Fl v
 output the version and exit.
+.It Fl S
+disable sorting.
+.Pp
+Instead of filtered and sorted by score, choices will only be filtered.
```
mike-burns: This is missing a verb. One of these two should do it:

* Instead of being filtered and sorted by score, choices will only be filtered.
* Instead of filtering and sorting by score, only filter the choices.

Maybe we can reverse the focus:

* Only filter the choices instead of additionally sorting by score.

mike-burns:
LGTM. I like how simple the code is.

man/man1/pick.1
calleerlandsson: I like the reversed one! Changed in 1f285e8.

calleerlandsson:
Thanks @mike-burns. I'll merge this.

PR #110, opened 2016-04-25
Fix standout rendering of last line

mptre:
Emitting the `clr_eos` capability will clear all columns from the
current cursor position (inclusive) to the end of the screen. If the
last line was selected the last column would therefor also be cleared
including the standout. The solution is simply to clear the screen prior
redrawing the interface.


calleerlandsson:
That makes sense. Thanks, @mptre!

PR #112, opened 2016-04-30
Change get_key to return the read key

mptre:
Since the number of recognized keys is limited make the `get_key'
function return a integer constant representing the read key. This
removes the need to cast the key buffer to an integer which broke strict
aliasing.


calleerlandsson:
This looks great to me @mptre! Thanks for taking the time to work on this.

PR #113, opened 2016-05-02
fix exit status of awk

prahlad37:
field() is always returning exit status 1, because END will be executed always, even after exit command inside body of awk.

ref: https://www.gnu.org/software/gawk/manual/html_node/Exit-Statement.html

According to above mentioned reference, exit in the END will use default success status, or use the exit status from the inner exit.
If  "exit:" is not defined in the test case then default EXIT_SUCCESS(0) is used and echo 0 will make the default exit code to 0.
If "exit:" is defined then exit 1 from the body will set the exit code to 1(EXIT_FAILURE) and same is used by the exit in END and the echo 0 will be skipped. 
Signed-off-by: Prahlad V <prahlad.eee@gmail.com>

### DIFF
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -2,8 +2,8 @@
 
 field() {
   awk '
-  /^'$1':/ { sub(/^[^:]+: */, ""); print; exit 0 }
-  END { exit 1 }'
+  /^'$1':/ { sub(/^[^:]+: */, ""); print; found=1 }
+  END { exit !found }'
 }
 
 input() {


calleerlandsson:
Hey @prahlad37, thanks for contributing!

I'm not super familiar with awk(1) but this looks good to me. @mike-burns, @mptre would you mind looking at this?

mike-burns:
Indeed, to quote POSIX:

> If an awk program consists of only actions with the pattern BEGIN, and the BEGIN action contains no getline function, awk shall exit without reading its input when the last statement in the last BEGIN action is executed. If an awk program consists of only actions with the pattern END or only actions with the patterns BEGIN and END, the input shall be read before the statements in the END actions are executed.

So this fix is correct.

mptre:
Your interpretation of the behavior is correct. However inverting the
meaning of the exit code does not make sense to me since the or
operator is used to reflected a fallback value. I propose the following
change instead:

```diff
diff --git tests/test.sh tests/test.sh
index 3de583a..0b13640 100755
--- tests/test.sh
+++ tests/test.sh
@@ -5,2 +5,2 @@ field() {
-  /^'$1':/ { sub(/^[^:]+: */, ""); print; exit 0 }
-  END { exit 1 }'
+  /^'$1':/ { sub(/^[^:]+: */, ""); print; found=1 }
+  END { exit !found }'
```

prahlad37:
using a variable to decide the exit status will be a better option.
Changes are updated.

mptre:
On Tue, May 03, 2016 at 06:09:47AM -0700, prahlad wrote:
> using a variable to decide the exit status will be a better option.
> Changes are updated.
> 
> ---
> You are receiving this because you were mentioned.
> https://github.com/thoughtbot/pick/pull/113#issuecomment-216521784

Looks good!

calleerlandsson:
Thanks for figuring this out together @prahlad37 and @mptre. Great work! I'll merge this.

calleerlandsson:
Merged as 0ab3f07de2cfcde81c5378643a3b579be0b00d92.

PR #114, opened 2016-05-02
Revert unintentional escape macro change

mptre:
Regression from e79f95b. ESCAPE is not a recognized key but instead used
in the `count_printable` function in order to recognize a escape
character.


calleerlandsson:
Thanks @mptre, I'll merge this.

Issue #116, opened 2016-05-04
Test fail


ScoreUnder:
Compiling from e79f95bbd208c580ea424848f64b34fb1ded10c4.
When running `make check`, the first test fails and the rest don't get tested.

test-suite.log:
```
==================================
   pick 1.3.0: ./test-suite.log
==================================

# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: tests/test.sh
===================

./tests/01.in: wrong exit code
test: read: Input/output error
FAIL tests/test.sh (exit status: 1)

```

Pick itself works fine, from what I can tell.

----

For the record, I've checked this against a fresh install of Arch under docker (pritunl/archlinux), and I get the same result.

mptre:
Could you try applying the following patch[1] locally and see if the
problem still remains?

[1] https://github.com/thoughtbot/pick/pull/113

ScoreUnder:
After applying the patch, `make check` still fails with the same error, but this time on test 03.

`git diff` after applying the patch:
```diff
diff --git a/tests/test.sh b/tests/test.sh
index 3de583a..0b13640 100755
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -2,8 +2,8 @@
 
 field() {
   awk '
-  /^'$1':/ { sub(/^[^:]+: */, ""); print; exit 0 }
-  END { exit 1 }'
+  /^'$1':/ { sub(/^[^:]+: */, ""); print; found=1 }
+  END { exit !found }'
 }
 
 input() {
```

mptre:
What if you apply the following patch:

```diff
diff --git tests/test.c tests/test.c
index 2402e5c..33b4291 100644
--- tests/test.c
+++ tests/test.c
@@ -70,7 +70,7 @@ parseinput(const char *s)
 				return;
 			switch (*s) {
 			case 'b':
-				c = 0x08;
+				c = 0x7F;
 				break;
 			case 'e':
 				c = 0x1B;
```

ScoreUnder:
That gets it a lot farther. It now fails on test 8. 

It's the same error code as before:

```
./tests/08.in: wrong exit code
test: read: Input/output error
FAIL tests/test.sh (exit status: 1)
```

mptre:
On Thu, May 05, 2016 at 10:42:10AM -0700, Score_Under wrote:
[...]

PR #118, opened 2016-05-14
Avoid EIO while testing on linux

ScoreUnder:
Relates to #116


calleerlandsson:
Thanks for putting time and effort in to this, @ScoreUnder! I'll merge this.

PR #119, opened 2016-05-16
Fix tcsetattr timing issue

mptre:
The input was being written prior the tcsetattr call in the pick
process which caused line editing to alter the input.

While at it: sort variables by size according to style(9).

Partial solution to #116.


calleerlandsson:
Thanks @mptre, this looks and works great. I'll merge this.

PR #120, opened 2016-05-21
Cleanup

mptre:
Time for another round of cleanups. The commits should hopefully be
self-explanatory. Other than that, here's two things that I would like to see
happening:

- Since both `choices` and `input` are data-structures that live through-out
  the whole execution of pick I would rather omit the free-ing and let the OS
  handle it instead. This is a common pattern among other BSD-commands.

- I find the following line in the `main` function confusing:

  ```c
  use_alternate_screen = getenv("VIM") == NULL;
  ```

  It should not the alternate screen if executed inside Vim(?). Wouldn't be
  better if the pick.vim plugin explicitly made of the `-X` option? If not:
  a comment would be useful since the behavior seems very specific.

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -5,7 +5,6 @@
 #include <sys/ttydefaults.h>
 #include <ctype.h>
 #include <err.h>
-#include <limits.h>
 #include <locale.h>
 #include <signal.h>
 #include <stdio.h>
--- a/src/pick.c
+++ b/src/pick.c
@@ -330,7 +330,8 @@ selected_choice(void)
 		switch (key) {
 		case ENTER:
 			if (visible_choices_count > 0) {
-				if (selection >= 0 && selection < (ssize_t)choices.length)
+				if (selection >= 0
+				    && selection < (ssize_t)choices.length)
 					return &choices.v[selection];
 				else
 					return NULL;
--- a/src/pick.c
+++ b/src/pick.c
@@ -158,8 +158,8 @@ main(int argc, char **argv)
 	if (query == NULL) {
 		query_size = 64;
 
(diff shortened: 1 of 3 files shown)


calleerlandsson:
This looks great, @mptre! I'll take a closer look at the code when I get some time over on Friday.

I agree that the check for `VIM` adds unnecessary complexity to pick and I think we should update pick.vim to use `-X` explicitly.

Since the OS should free all allocated memory of a process when it terminates I think omitting the `free`s should be fine. The reason I included them in the first place was because I thought it was best practice. Do you have any recommended reading on the practice of not explicitly `free`ing? @mike-burns What are your thoughts on the freeing?

My general opinion is that less code is always better.

mike-burns:
> @mike-burns What are your thoughts on the freeing?

Omitting `free` when terminating a program is fine, but remember to add them back when you refactor. I find it easier to leave them in, even at the end, for that reason.

teoljungberg:
@calleerlandsson I updated pick.vim with that suggestion in https://github.com/thoughtbot/pick.vim/pull/9

calleerlandsson:
I've had some more time to look at this now and the code looks great! It also runs well on my system and all tests pass. I'll merge this. Thanks for keeping up the amazing work, @mptre!

I think it is time to cut a release soon. What are your thoughts?

@mike-burns Thanks for your input.

@teoljungberg Thanks for preparing for the upcoming change!

mptre:
I just submitted two small patches. If possible, I would like them to be
included before doing a release.

calleerlandsson:
@mptre Sure thing! Let's merge the PRs and run the current master for a week and release on Friday.

PR #123, opened 2016-05-27
Remove altscreen condition if invoked from Vim

mptre:
The pick.vim plugin now makes explicit use of the `-X' option. As
discussed in #120.


calleerlandsson:
Thanks, @mptre!

This changes the behavior of defaulting `use_alternate_screen ` to `1` to defaulting to `0`. Should we initialize `use_alternate_screen` to `1` on line 96?

mptre:
Nice catch! Fixed and rebased.

PR #124, opened 2016-05-27
Pledge on OpenBSD

mptre:
Thoughts @mike-burns?

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -22,6 +22,9 @@
 /* Define if ncursesw is available */
 #undef HAVE_NCURSESW_H
 
+/* Define to 1 if you have the `pledge' function. */
+#undef HAVE_PLEDGE
+
 /* Define to 1 if your system has a GNU libc compatible `realloc' function,
    and to 0 otherwise. */
 #undef HAVE_REALLOC
--- a/configure.ac
+++ b/configure.ac
@@ -8,7 +8,7 @@ AC_TYPE_SIZE_T
 AC_TYPE_SSIZE_T
 AC_FUNC_MALLOC
 AC_FUNC_REALLOC
-AC_CHECK_FUNCS([memmove reallocarray setlocale strdup])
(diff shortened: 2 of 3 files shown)


mike-burns:
Rad! I'll have to try it later, but quickly: how did you decide on those promises?

mptre:
Start with the empty string. In case of a crash, `dmesg | grep pick`
will tell you which promise is missing, or examine the core-dump using
gdb in order to understand why.

mike-burns:
I played with it and was able to re-pledge before the user can give input. I used ktrace(1) to help:

```
 make && echo "1\n2\n3\n4" | ktrace ./src/pick ; ktrace -C ; kdump > trace && tail trace
```

In which I learned that, by pledging where you pledged:

* `cpath` and `wpath` are needed to open `/dev/tty`.
* `rpath` is needed to open `/usr/share/locale/UTF-8/LC_CTYPE` and then `~/.terminfo`.
* `tty` is needed for the ioctl(2) requesting `TIOCGETA` (and likely others).

So I re-pledged at 177, just before calling `selected_choice()`:

```c
if (pledge("stdio tty", NULL) == -1)
	err(1, "pledge");
```

There, `tty` is needed because `restore_tty()` is called afterward, which uses ioctl(2) requesting `TIOCSETA` (and likely others).

What do you think of pledging twice: once where you did, and again just before `selected_choice()`?

mptre:
> What do you think of pledging twice: once where you did, and again just before `selected_choice()`?

Aiming for the minimal subset of needed promises makes perfect sense.
Good catch!

I updated the patch with your changes and improved the commit-message.

mike-burns:
Nice. Looks good to merge.

calleerlandsson:
That's a cool feature of OpenBSD. Thanks for adding this, @mptre! Merged as 1bf6fb42b5220a217aba9c0e8f26cd9459011441.

PR #126, opened 2016-05-29
Run tests with MALLOC_OPTIONS=S OpenBSD

mptre:
Enable malloc.conf(5) options on OpenBSD which will abort the pick process when
reading/writing out-of-bounds or accessing already freed memory. A core-file is
created on failure and can be inspected using gdb(1):

```
gdb -c pick.core src/pick
```

### DIFF
--- a/tests/test.c
+++ b/tests/test.c
@@ -116,6 +116,12 @@ child(int master, int slave)
 	if (ioctl(slave, TIOCSCTTY, NULL) < 0)
 		err(1, "ioctl");
 
+	/* Enable malloc.conf(5) options on OpenBSD which will abort the pick
+	 * process when reading/writing out-of-bounds or accessing already freed
+	 * memory. */
+	if (setenv("MALLOC_OPTIONS", "S", 0) < 0)
+		err(1, "setenv");
+
 	execvp(argv[0], (char *const *)argv);
 	e = errno;
 	err(1, "%s", argv[0]);


tests/test.c
```diff
@@ -116,6 +116,12 @@ child(int master, int slave)
 	if (ioctl(slave, TIOCSCTTY, NULL) < 0)
 		err(1, "ioctl");
 
+	/* Enable malloc.conf(5) options on OpenBSD which will abort the pick
+	 * process when reading/writing out-of-bounds or accessing already freed
+	 * memory. */
+	if (setenv("MALLOC_OPTIONS", "S", 0) < 0)
```
mike-burns: This can alternatively be done with the global `malloc_options` char pointer; see [`usr.sbin/httpd/patterns/patterns-tester.c`](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/regress/usr.sbin/httpd/patterns/patterns-tester.c?rev=1.1&content-type=text/x-cvsweb-markup) for an example (lines 25 and 66).

mike-burns:
Looks good to me.

I had to read `malloc.c`, but `S` = `FGJP`.

I have no feelings regarding the env var vs. global variable. Global might be faster, though.

tests/test.c
mptre: I used the environment variable since the current process is replaced
when calling `execvp`.

mike-burns: Ah right, that makes sense!

mptre:
> Looks good to me.

Thanks for taking the time to review.

> I had to read `malloc.c`, but `S` = `FGJP`.

Me too, concluded that it looked like reasonable defaults for the
use-case. 

> I have no feelings regarding the env var vs. global variable. Global might be faster, though.

See comment above.

calleerlandsson:
Looks great to me. I'll merge this.

mptre:
Glad to inform that this change found 485b436.

PR #128, opened 2016-06-05
Undefine __dead before redefining it if necessary

fornwall:
This fixes building for Android.

### DIFF
--- a/src/compat.h
+++ b/src/compat.h
@@ -10,10 +10,12 @@
 #endif /* __FreeBSD__ */
 
 #if defined(__linux__) || defined(__CYGWIN__)
+#ifndef __dead
 #ifdef __GNUC__
 #define __dead		__attribute__((__noreturn__))
 #else
 #define __dead
+#endif
 #endif
 #endif /* __linux__ || __CYGWIN__ */
 


calleerlandsson:
Hey @fornwall, thanks for contributing!

What is `__dead` defined as for Android? Would it make sense to instead keep the original definition? If so, maybe we should wrap the whole `__dead` combat section in an `#ifndef __dead`?

fornwall:
In `<sys/cdefs.h>`:
```
/* Delete pseudo-keywords wherever they are not available or needed. */
#ifndef __dead
#define __dead
#define __pure
#endif
```

While keeping that empty define would at least not break the build here, perhaps redefining is safer and better in general.

calleerlandsson:
Thanks for finding that!

My understanding is that, if defined, `__dead` will let the compiler optimize functions that never return. Therefor, I think we want to preserve whatever value `__dead` has.

If it is not defined, however, we need to define it in order `pick.c` to compile.

I might be getting this completely wrong, though!

Maybe @mike-burns or @mptre know more about this?

mptre:
> My understanding is that, if defined, `__dead` will let the compiler optimize functions that never return. Therefor, I think we want to preserve whatever value `__dead` has.

Correct.

> If it is not defined, however, we need to define it in order `pick.c` to compile.

Correct.

fornwall:
Ok, should I change the pull request to wrap the whole `__dead`-mangling block (line 8 to 18 in `compat.h`) with `ifndef __dead`?

calleerlandsson:
@fornwall How about we remove line 15 and 16 and add this above `#ifndef HAVE_REALLOCARRAY`:

```
#ifndef __dead
#define __dead
#endif

```

That way we can explicitly define `__dead` to be what we want it to be on systems and compilers we know and let it fall back to whatever it defaults to on other systems that have it and to nothing on systems that don't.

fornwall:
Ah, I should have posted the original problem:

    compat.h:14:0: error: "__dead" redefined

So the problem is that [line 14](https://github.com/thoughtbot/pick/blob/master/src/compat.h#L14) tries to redefine `__dead` without checking if it is already defined.

calleerlandsson:
Oh, right, that makes sense! Let's wrap the whole thing.

fornwall:
I've updated this pull request with a `ifndef __dead` wrapping.

calleerlandsson:
Thanks @fornwall, looks great! I'll merge this.

calleerlandsson:
Merged as 6006a9b.

PR #132, opened 2016-06-10
Reduce flickering

### LINKED ISSUE #131: Visual regression in 1.4
While using `pick` under `tmux` - the screen flickers while traversing the results.

out

I'm running pick 1.4.0 and tmux 2.1.

calleerlandsson:
The clearing of the screen introduced in 72239f0 made the screen flicker
for users in tmux. Possibly because tmux might take longer time to draw
than terminal emulators.

Instead of clearing the whole screen before printing the choices and
query, clear from the line below the last choice to the end of the
screen before printing the query. This only can and needs to be done if
the visible choices and the query do not completely fill the screen.

This might resolve #131.

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -476,6 +476,10 @@ selected_choice(void)
 		tty_putp(cursor_invisible);
 
 		visible_choices_count = print_choices(selection);
+		if (visible_choices_count < lines - 1) {
+			tty_putc('\n');
+			tty_putp(clr_eos);
+		}
 		if (cursor_position >= scroll + columns)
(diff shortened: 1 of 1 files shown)


calleerlandsson:
@mptre how does this look to you?

mptre:
This change breaks the output when not using the alternate screen and
the terminal needs to scroll in order fit all choices unfortunately.

calleerlandsson:
Oh yeah, thanks for noticing that @mptre! Let's close this in favor of #133.

PR #133, opened 2016-06-12
Refactor print logic

mptre:
An attempt to solve #131 turned into a bigger change than expected, for better IMHO. See each commit messages for more info. Would really like to get feedback from all of you.

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -80,7 +80,6 @@ static int			 print_choices(int);
 static int			 get_key(char *, size_t, size_t *);
 static int			 tty_getc(void);
 static void			 delete_between(char *, size_t, size_t, size_t);
-static void			 print_query(char *, size_t, size_t, size_t);
 static int			 choicecmp(const void *, const void *);
 static int			 isu8cont(unsigned char);
 static int			 isu8start(unsigned char);
@@ -289,7 +288,7 @@ put_choice(const struct choice *choice)
 const struct choice *
 selected_choice(void)
 {
-	size_t		cursor_position, length, query_length, scroll;
+	size_t		cursor_position, i, length, query_length, scroll = 0;
 	char		buf[6];
 	int		key, selection = 0, visible_choices_count;
 	int		word_position;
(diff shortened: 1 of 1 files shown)


src/pick.c
```diff
@@ -648,93 +633,79 @@ restore_tty(void)
...
+			if (string[i] >= '@' && string[i] <= '~')
+				in_esc_seq = 0;
+		} else if (i > 0 && string[i - 1] == ESCAPE
+		    && string[i] == '[') {
+			in_esc_seq = 1;
+			non_printable = 2;
+		}
+
+		c = string[i];
+		switch (c) {
```
calleerlandsson: Since there are only two cases, do you think an if statement would make this clearer?

```
if (c == '\0') {
	/*
 	 * A null character will be present prior the
 	 * terminating null character if descriptions is
 	 * enabled.
 	 */
 	c = ' ';
 	col++;
} else if (!isu8cont(c))
	col++;
```

src/pick.c
```diff
@@ -648,93 +633,79 @@ restore_tty(void)
...
-	for (i = 0; i < length && i < columns; ++i) {
-		if (tty_putc(string[i]) == EOF)
+	for (col = i = 0; i < length && col < columns; i++) {
+		if (i == (size_t)underline_start)
+			tty_putp(enter_underline_mode);
+
+		/*
+		 * Count the number of outputted ANSI escape sequences
+		 * in order to adjust the column count since they do not
+		 * occupy any screen real-estate.
```
calleerlandsson: I think this is a great solution!

calleerlandsson:
Hey @mptre, thanks so much for this! It looks, builds, and runs great on my system. I had one small question. Other than that this looks good to merge!

@teoljungberg Would you like to confirm that this resolves the flickering issues in tmux on your system?

teoljungberg:
@calleerlandsson Indeed it does!

calleerlandsson:
Thanks @teoljungberg.

teoljungberg:
Thank _you_ two for working on this!

src/pick.c
mptre: Totally valid point. However, I plan on fixing output of tabs later
which is currently broken. Then the number of conditions is no longer
binary, therefore the pro-active switch.

mptre:
I sneaked in a minor cleanup, let me know what you think. Other that
than that, it's ready to be merged.

calleerlandsson:
Looks great, @mptre! Could you rebase this on the current master and squash in the mixup commit?

mptre:
Done!

calleerlandsson:
Thanks! I'll merge this.

PR #134, opened 2016-06-16
Update Homebrew installation instructions

calleerlandsson:
Since Pick was added to Homebrew core in Homebrew/homebrew-core#1977,
adding the thoughtbot tap before installing with `brew` is no longer
needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,6 @@ apt-get install pick
 ### Mac OS X via Homebrew
 
 ```sh
-brew tap thoughtbot/formulae
 brew install pick
 ```
 


mptre:
Congratulations on the core-promotion! The test[1] should be running `make
check` by now.

[1] https://github.com/thoughtbot/pick/blob/master/homebrew/pick.rb.in#L13-L15

calleerlandsson:
Thanks, @mptre! That's a great point, I'll mention it on the homebrew-core PR.

calleerlandsson:
I'll merge this.

PR #135, opened 2016-06-16
Remove Homebrew release tasks

calleerlandsson:
The Pick Homebrew formula is now maintained within the
Homebrew/homebrew-core project and updating it can no longer be done
automatically.


calleerlandsson:
The Arch Linux release tasks do not work very well so I'll follow up with a PR removing those as well.

I think maintaining system packages outside of this repo creates a clear separation of concerns.

mike-burns:
LGTM.

calleerlandsson:
Thanks @mike-burns. I'll merge this.

PR #142, opened 2016-06-28
Improve search performance by not wasting cycles

mptre:
Using pick with a large number of choices is slow. The threshold on my
computer is somewhere around 10k lines of input. Benchmarking the
filtering revealed that the scoring is the bottleneck. This is no
surprise since the current search algorithm is an `O(n^2)` operation in
worst case. The second most time consuming task is the sorting of the
choices. However, the time taken for `qsort(3)` is consistent and several
order of magnitude less time consuming than the scoring.

When the user inputs a query longer greater than 1 character we
currently waste lots of cycles by performing a search every time a new
character is received. Even if the current query is outdated since the
user already has typed more character yet to be processed. Therefore,
abort the search if there is unprocessed user input available. This is
done by performing a non-blocking `poll(2)` call on the user input file
descriptor. The check for new input is done on every 50:th iteration.
The number is empirically chosen and close to the number of lines in my
terminal. There's no guarantee that this number is optimal for everyone.
Experimenting with the number of different computers is therefore of
interest.

This change gives the notion of a responsive interface even when the
input is large. Running the following command prior and after this
change should result in a noticeable speed-up:

```sh
$ find ~ | pick
```

I think this a decent trade-off in terms of gained performance vs.
complexity.

Tested on OpenBSD and Ubuntu. However, it's currently **not working on OS
X** since `pfd.revents & POLLNVAL`[1] is always true. Any help to resolve
this issue would be much appreciated. 

[1] The corresponding file descriptor is invalid

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -6,6 +6,7 @@
 #include <ctype.h>
 #include <err.h>
 #include <locale.h>
+#include <poll.h>
 #include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -471,13 +472,35 @@ selected_choice(void)
 	}
 }
 
+/*
+ * Filter choices using the current query while there is no new user input
+ * available.
+ */
 void
 filter_choices(void)
 {
-	int	i;
+	struct pollfd	pfd;
+	size_t		i;
+	int		ready;
 
-	for (i = 0; i < (ssize_t)choices.length; ++i)
+	for (i = 0; i < choices.length; i++) {
 		score(&choices.v[i]);
+
+		/*
+		 * Regularly check if there is any new user input available. If
+		 * true, abort filtering since the currently used query is
+		 * outdated. This improves the performance when the cardinality
+		 * of the choices is large.
+		 */
+		if (i % 50 == 0) {
+			pfd.fd = fileno(tty_in);
+			pfd.events = POLLIN;
+			if ((ready = poll(&pfd, 1, 0)) == -1)
+				err(1, "poll");
+			else if (ready == 1 && pfd.revents & (POLLIN | POLLHUP))
+				break;
+		}
+	}
 
 	qsort(choices.v, choices.length, sizeof(struct choice), choicecmp);
 }
--- a/src/pick.c
+++ b/src/pick.c
@@ -481,7 +481,7 @@ filter_choices(void)
 {
 	struct pollfd	pfd;
 	size_t		i;
-	int		ready;
+	int		nready;
 
 	for (i = 0; i < choices.length; i++) {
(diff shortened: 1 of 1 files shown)


calleerlandsson:
This is a great improvement! Thanks, @mptre!

Some thoughts and questions:

* Running this on macOS, I don't notice any errors or difference in behavior to master. Is this how it is currently not working or have you seen errors? If it "falls back" to the previous behavior, I could imagine merging this for the benefit of other platforms while we figure out the macOS specific issues.

* Do you think there are any straight forward improvements we could make to the search algorithm?

I'll dig deeper in the macOS issues whenever I get a chance.

mptre:
> * Running this on macOS, I don't notice any errors or difference in behavior to master. Is this how it is currently not working or have you seen errors? If it "falls back" to the previous behavior, I could imagine merging this for the benefit of other platforms while we figure out the macOS specific issues.

The behavior on macOS is kept the same but won't be able to benefit from
this change until the poll-issue is resolved.

> * Do you think there are any straight forward improvements we could make to the search algorithm?

Reducing the time complexity of the current shortest substring algorithm
would require some kind of pre-computation[1]. I don't think such
algorithms is suited for pick, I like current one for its simplicity and
decent running time. I have never been bothered by the performance of
the current algorithm for my day-to-day tasks involving pick. But the
poll optimization was such low-hanging fruit.

I've also been experimenting with pruning the number of choices to
evaluate on each `find_choices` iteration. The basis of the optimization
relies on following fact: if a choice `C` does not contain the query `Q`
it will never contain the query `Q'` if `|Q'| > |Q|` and `Q` is a subset
of `Q'`. This gave a noticeable improvement when the number of choices
is high. But in conjunction with poll-optimization it becomes useless
under perfect circumstances since the poll-optimization will run the
scoring-function first when the user has stopped typing and at that
point only a few choices has already been scored and can be pruned.

> I'll dig deeper in the macOS issues whenever I get a chance.

Great!

[1] https://en.wikipedia.org/wiki/Wagner–Fischer_algorithm

mptre:
Some interesting findings on the Darwin related problems regarding poll(2):

- https://github.com/tmux/tmux/blob/5658b628b9bf1c1e0bd5856736332ce8b9c51517/osdep-darwin.c#L73-L76

- https://github.com/tmux/tmux/blob/c84ed36719f8bfc72ef52e8e0a43003464de587e/CHANGES#L1399

calleerlandsson:
Yeah, although old, these comments are a bit disconcerting.

I think this looks good to merge for now, though. Let's figure out macOS support separately.

mptre:
Merged as 1ba7398.

Issue #138, opened 2016-06-20
Choices are limited to one screen


dstd:
Am I doing something wrong, or it's a limitation of `pick` – it shows only one screen of options.
So that the only way to select the items that didn't fit into the screen is to adjust the query?

// mac, homebrew, pick 1.4.0

kylebebak:
Same for me, OSX, pick 1.3.0

Being able to scroll through more than one page of results would be fantastic

mptre:
I've been missing this feature as well. I'm planning on trying to
implement this soon.

calleerlandsson:
Thanks for the feedback, everyone! I also think this would be a great addition.

mptre:
#144 has now been merged into the master branch and therefore assume this issue can be closed?

calleerlandsson:
Definitely. Thanks, @mptre!

PR #144, opened 2016-07-02
Add vertical scrolling support

mptre:
Solution to #138.

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -285,7 +285,7 @@ const struct choice *
 selected_choice(void)
 {
 	size_t	cursor_position, i, length, query_length;
-	size_t	scroll = 0;
+	size_t	xscroll = 0;
 	char	buf[6];
 	int	choices_count, key, word_position;
 	int	selection = 0;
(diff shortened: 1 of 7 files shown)


src/pick.c
```diff
@@ -463,6 +470,7 @@ selected_choice(void)
 			filter_choices();
 			selection = 0;
 		}
+		yscroll = 0;
```
calleerlandsson: As far as I understand, we only want to do this when the query is changed or put differently; whenever we do the following:

```
filter_choices();
selection = 0;
```

Reading through the code, I notice this bit being duplicated in all cases where the query is changed.

In all other cases we also want to skip resetting `yscroll`. For example, in the left and right arrow key cases. Currently after advancing `yscroll`, pressing left or right arrow will reset it to `0`.

I see two possible solutions to this:

- We could add `yscroll = 0;` wherever the above bit is duplicated. This would introduce more duplication but would fix the bug.
- Alternatively, we could extract the call to `filter_choices` and resetting the `selection` to just above this line and add `continue` statements to all branches that shouldn't go through the reset.

calleerlandsson:
Awesome, thanks @mptre! I've added one comment above.

src/pick.c
mptre: Nice catch! I ended up begin explicit and resetting `yscroll` when
appropriate. See ab2fe6f.

calleerlandsson:
Looks good to merge. Thanks for working on this!

tests/test.c
```diff
@@ -116,6 +117,12 @@ child(int master, int slave)
 	if (ioctl(slave, TIOCSCTTY, NULL) < 0)
 		err(1, "ioctl");
 
+	/* Set window size to known dimensions, necessary in order to deduct
```
mike-burns: s/deduct/deduce/

mptre: Ty

mptre:
Merged as dece164 and ebf87b4.

PR #148, opened 2016-07-10
Rewrite tests

mptre:
Work in progress of rewriting the tests. Notable changes:

- Using the Automake Log Compiler[1] feature improves the output of
  `make check` and parallelism can be utilized (`make check -jN`)

- Rename `tests/test.c -> tests/pick-test.c` and adopt `style(9)`

- Change test-case file format, everything is now a field

- The rewrite highlighted an issue with `OXTABS` on Linux

- Tested on OpenBSD, Ubuntu and OS X

Question for @mike-burns and all others: can the `TESTS` variable[2] be
initialized using a wildcard `tests/*.t` without using any GNU-ism?

[1] https://www.gnu.org/software/automake/manual/html_node/Parallel-Test-Harness.html
[2] https://github.com/thoughtbot/pick/blob/2ccace20a9ddfe44fc1d0b1e11e4373a089dde23/Makefile.am#L10-L14


mike-burns:
Wildcards: no, sadly not. They're philosophically opposed: https://www.gnu.org/software/automake/manual/html_node/Wildcards.html

tests/pick-test.c
```diff
@@ -0,0 +1,220 @@
...
+			}
+		} else if (c == ' ') {
+			continue;
+		}
+		buf[len++] = c;
+
+		if (size <= len) {
+			if ((buf = realloc(buf, 2*size)) == NULL)
+				err(1, NULL);
+			size *= 2;
```
mike-burns: Totally doesn't matter for this test script, because we control all the input, but this could technically lead to overflow. Not saying you should do anything about this though.

tests/pick-test.c
```diff
@@ -0,0 +1,220 @@
...
+child(int master, int slave)
+{
+	struct winsize	ws;
+	int		fd;
+
+	close(master);
+
+	/* Disconnect the controlling tty. */
+	if ((fd = open("/dev/tty", O_RDWR|O_NOCTTY)) == -1)
+		err(1, "open");
```
mike-burns: Could be `err(1, NULL)` (and elsewhere in here).

mike-burns:
LGTM overall.

tests/pick-test.c
mptre: I've changed the error message to include the path for all open related
errors[1]. Do you still advocate on passing `NULL` as the message on
failure?

[1] 27991aa3cd6614aa7c2df8b17210028b94f39dd0

tests/pick-test.c
mptre: Correct. I didn't bother including `src/compat.c` to access
`reallocarray(3)`. If we want to play it safe, how about:

```c
if (size > SIZE_MAX/2)
	errx(1, "overflow");
else if ((buf = realloc(buf, 2*size)) == NULL)
	err(1, NULL);
```

If not, maybe leaving a comment for now will do?

tests/pick-test.c
mike-burns: That path is a great improvement -- keep that!

tests/pick-test.c
mike-burns: I like the comment solution -- shows that we are aware of it and will happily accept fixes for it should it ever become a problem.

mptre: Fixed in b766d62.

mptre:
Updated documentation and squashed, merging for now.

PR #149, opened 2016-07-17
Add support for page keys

mptre:
I use `git-blame(1)` with pick, when blameing a long file I prefer being able to
view the line/chunk I want to blame without having to enter a query first. This
is tedious when the wanted line is below the first page of choices. Pressing
Page-Down/Page-Up will scroll a page (number of lines in your terminal) worth of
choices.


src/pick.c
```diff
@@ -446,6 +448,14 @@ selected_choice(void)
 			while (cursor_position < query_length
 			    && isu8cont(query[++cursor_position]));
 			break;
+		case PAGE_DOWN:
+			if (yscroll + lines - 1 < choices_count)
```
calleerlandsson: How does this handle the case where there are still choices below the selection and possibly below the scroll, but not more than `lines - 1`.

If a full page can't be scrolled (if this condition is false), I'd expect it to scroll to and select the last choice. What are your thoughts?

calleerlandsson:
Great idea @mptre! The code looks good, I just had one question about behavior.

mptre:
Pushed two commits:

- Found a bug related to clearing the screen introduced while adding
  support for vertical scrolling. @teoljungberg: could you try this
  branch in order to verify that this change does not introduce any
  flickering on your setup?

- Page up works when less than a full page has been scrolled which is
  the expected behavior IMHO.

src/pick.c
mptre: Good point! I just changed the page up behavior to always scroll even
when less than a page has been scrolled, in this scenario the vertical
scroll is removed and the first choice is selected. For page down, we
could either:

- Scroll and only show the last choice

- Select the last choice but do not scroll any further. This is somewhat
  more consistent to the current page up behavior and how mutt(1)
  behaves as well.

I vote for the later, what's your take?

calleerlandsson: That's a great improvement!

When I try it out locally it still feels a bit unnatural, though. For example, if I select the third choice by pressing the down arrow key twice and then press page up, nothing happens. What I'd expect is for the first choice to be selected. Another example is if I select the last choice in a list with more choices than would fit two pages, and then press page up it will advance the selection by two pages - 1.

I think what confuses me is that it is not immediately clear to me what the page up/down keys are controlling. In the case of the arrow keys, it is clearly controlling the selection and then the choices are scrolled to always show the selected one.

What do you think about letting the page up/down keys also control the selection by advancing it by a `lines` number of choices and then scrolling the choices to show the selection?

If `selection < lines` then it could `selection = 0` and when `choices_count - selection < lines` then it could `selection = choices_count`.

mptre: Great input! I agree, the current behaviour is confusing. I followed
your advice and made the page down/up move the selection one page
down/up, counting from the current selection. If less choices than a
full page is available the first/last choice is selected. The man-page
is updated, do you think it describes the behaviour clear enough?

calleerlandsson: The new behavior works great and the description sounds good!

calleerlandsson:
I think this looks good to merge, @mptre. Thanks for working on this!

mptre:
Merged as bd5819f and d3b1206.

teoljungberg:
@mptre Sorry for being late to the party, I've been on vacation for the last few weeks.

I cloned down this branch(or the recent master specifically) and this worked just fine on my setup. Great work!

mptre:
@teoljungberg thanks for taking the time to try out this change!

PR #150, opened 2016-07-30
Cleanup

mptre:
Time for another round of cleanups. As always, let me know what you
think.

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -104,7 +104,7 @@ int
 main(int argc, char **argv)
 {
 	const struct choice	*choice;
-	int			 option;
+	int			 c;
 	int			 output_description = 0;
 
 #ifdef HAVE_PLEDGE
@@ -114,8 +114,8 @@ main(int argc, char **argv)
 
 	setlocale(LC_CTYPE, "");
 
-	while ((option = getopt(argc, argv, "dhoq:SvxX")) != -1) {
-		switch (option) {
+	while ((c = getopt(argc, argv, "dhoq:SvxX")) != -1) {
+		switch (c) {
 		case 'd':
 			descriptions = 1;
 			break;
--- a/src/pick.c
+++ b/src/pick.c
@@ -95,7 +95,7 @@ static struct {
 static FILE		*tty_in, *tty_out;
 static char		*query;
 static size_t		 query_size;
-static int		 descriptions;
+static int		 descriptions, choices_lines;
 static int		 sort = 1;
 static int		 use_alternate_screen = 1;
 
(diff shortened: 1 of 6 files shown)


mptre:
I thought this was worth sharing in order to get an idea of what the
latest optimizations (4c13591 and 1ba7398) means in practice. The
`ports.out` contains the name of all 135987 files in my `/usr/ports`
directory.

Before:

pick-old

After:

pick-new

src/pick.c
```diff
@@ -460,7 +459,8 @@ selected_choice(void)
 				yscroll = selection = 0;
 			break;
 		default:
-			if (!isu8start(buf[0]) && !isprint(buf[0]))
+			if (length == 0
```
calleerlandsson: I agree that zeroing out the input buffer isn't necessary. Thanks for catching that!

I'm not sure how the change to this condition is related to this. Could you elaborate on this? When is `length` zero?

src/pick.c
```diff
@@ -179,7 +174,6 @@ main(int argc, char **argv)
 	}
 
 	free(choices.v);
-	free(input.string);
```
calleerlandsson: Since we're only ever freeing this at the end of successful execution, I agree that letting the OS do this is fine.

src/pick.c
```diff
@@ -95,7 +95,7 @@ static struct {
 static FILE		*tty_in, *tty_out;
 static char		*query;
 static size_t		 query_size;
-static int		 descriptions;
+static int		 descriptions, tty_lines;
```
calleerlandsson: I can't really tell the difference between `tty_lines` and `lines` just from this name.

Would something like `choice_lines` make it clearer that it is `lines` minus 1 for the query line?

src/pick.c
```diff
@@ -491,7 +491,7 @@ filter_choices(void)
 		 * outdated. This improves the performance when the cardinality
 		 * of the choices is large.
 		 */
-		if (i % 50 == 0) {
+		if (i > 0 && i % 50 == 0) {
```
calleerlandsson: Great catch!

src/pick.c
mptre: You're right, `length` will never be zero. Fixed in 02e4a4c.

src/pick.c
mptre: Great proposal, fixed in f52c184.

calleerlandsson:
@mptre This looks great to me! Rebase, squash and merge at will.

mptre:
Thanks, merged!

PR #151, opened 2016-08-14
Improve test code coverage

mptre:
Using `gcov(1)` revealed a couple of code paths not covered by the
tests. Some of the uncovered behavior can't be verified using the
current test framework but a potential crash can still be detected.

### DIFF
--- a/tests/36.t
+++ b/tests/36.t
@@ -1,0 +1,7 @@
+description: delete til cursor
+keys: abc \002 \025 \\n
+stdin:
+abc
+c
+stdout:
+c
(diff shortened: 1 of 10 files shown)


tests/36.t
```diff
@@ -0,0 +1,7 @@
+description: delete til cursor
+keys: abc \002 \025 \\n
```
calleerlandsson: I don't think we should hold back on merging this PR due to this but a thought for future PRs:

It's not obvious to me what keys are pressed in this spec. I wonder if we could clarify by using some kind of key character aliases in the `keys` such as `Ctrl-B` or `^B` to mean `\002`? If not, what do you think about adding support for comments in the spec runner so one could write:

```
keys: abc \002 \025 \\n -- Ctrl-B followed by Ctrl-U and Enter
```

calleerlandsson:
Thanks for testing and improving this @mptre. Great work! I think this looks ready to merge.

tests/36.t
mptre: I decided to remove the ability to use aliases in 5fc3844 due to the
following reasons:

- Less code

- There's a 1-1 mapping between the keys present in test-files and the
  ones declared in the `get_key` function.

I like the idea of comment support in order to add clarity. Something
simple as:

```sh
val=$(echo $val | sed 's/#.*//')
```

... in `tests/test.sh` will probably do, thoughts?

mptre:
Thanks, merged. See my comment above regarding comment support.

PR #153, opened 2016-08-16
Right-justify backslashes according to style(9)

mptre:
From `style(9)`:

> Right-justify the backslashes, as the resulting definition is easier to
> read.


calleerlandsson:
Looks good to me!

mptre:
Thanks, merged as cb2eb4c.

PR #154, opened 2016-08-21
Fix potential integer overflow

mptre:
I decide it was worth trying to fix this potential overflow. Thoughts?


calleerlandsson:
Looks great!

mptre:
Thanks, merged.

PR #155, opened 2016-08-30
Discard unrecognized escape sequences

mptre:
If the user inputs a unrecognized escape sequence longer than the size
of the input buffer, parts of the sequence would end up in the query.

Since the sequence should be ignored, there's no need to put the read
bytes in the input buffer.


calleerlandsson:
Good catch @mptre! The code looks good to me. Feel free to merge!

mptre:
Thanks, merged!

PR #156, opened 2016-09-01
Remove usage of save and restore cursor

mptre:
Usage of cursor save and restore capabilities is non-obvious. Moving the
cursor the appropriate number of lines/columns is easier to follow.

While at it, make the query cursor movement a single operation as
opposed of the current implementation which is linear with the number of
characters in the query.

I would appreciate if others could test this change in order to not
introduce any regression.

### DIFF
--- a/src/pick.c
+++ b/src/pick.c
@@ -272,7 +272,7 @@ eager_strpbrk(const char *string, const char *separators)
 const struct choice *
 selected_choice(void)
 {
-	size_t	cursor_position, i, length;
+	size_t	cursor_position, i, j, length;
 	size_t	xscroll = 0;
 	char	buf[6];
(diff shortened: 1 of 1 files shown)


calleerlandsson:
Hey @mptre, sorry for the late feedback. The code looks great and I agree that it is easier to follow. Thanks for working on this! Pick runs without regressions on my machine.

@teoljungberg Would you mind testing it in your tmux setup that had issues before?

teoljungberg:
@calleerlandsson Looks good after testing this out locally.

mptre:
Thanks for testing, I'm merging now. While at it, time for another
release?

calleerlandsson:
Thanks @mptre! Yes, I think it's time for a release. I'll merge #158 and release.

PR #161, opened 2016-09-20
pick-test cleanup and comment support

mptre:
As discussed in #151.

Comments may be placed anywhere on a line. This works since a leading
comment will be evaluated to:

```sh
eval "# comment goes here=''"
```

... which is an no-op. Trailing comments are stripped using a sh(1)
suffix substitute parameter.


tests/pick-test.c
```diff
@@ -12,9 +12,78 @@
...
+}
+
+static void
+usage(void)
+{
+	fprintf(stderr, "usage: pick-test [-k keys] [-- argument ...]\n");
+	exit(1);
+}
+
 static char *
```
calleerlandsson: Now that his is declared up top, can we drop `static` here?

mptre: Correct. However, it's common to repeat the keyword at the definition as
well. I might revert the change to `src/pick.c` later.

mptre:
Thanks for the feedback, merged!

PR #162, opened 2016-10-01
Cleanup

mptre:
Found an error while trying to run the tests on Linux. Other than that,
only minor changes.


calleerlandsson:
LGTM! I like the explicit `continue`s. They make what's happening obvious.

mptre:
Thanks, merged.

PR #163, opened 2016-10-01
Don't exit for some missing capabilities

mptre:
All terminal capabilities used are not critical for pick to behave
correctly, such as `{enter,exit}_ca_mode` and `cursor_{invisible,normal}`.
This change makes pick work in OpenBSD console where `TERM=v100`.


calleerlandsson:
Great idea! The code looks good to me.

mptre:
Thanks, merged.

PR #165, opened 2016-10-04
Travis

mptre:
This is an attempt to prevent regression like the one I managed to
introduce in 1.5.0, where the tests could not run on Linux. Travis is
capable of building on both Linux and macOS. Here's a first attempt to
configure Travis. However, I don't have the privilegies needed to enable
Travis, ping @mike-burns and @calleerlandsson.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+# Use trusty environment for a up-to-date version of autotools
+sudo: required
+dist: trusty
 language: c
 os:
   - linux
@@ -7,6 +10,6 @@ compiler:
   - gcc
 script:
   - ./autogen.sh
-  - ./configure
+  - ./configure || (cat config.log; exit 1)
   - make || exit 1
   - make check || (cat test-suite.log; exit 1)
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: c
+os:
+  - linux
+  - osx
+compiler:
+  - clang
+  - gcc
+script:
+  - ./autogen.sh
+  - ./configure
+  - make || exit 1
+  - make check || (cat test-suite.log; exit 1)
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,7 +16,7 @@ TESTS=tests/01.t tests/02.t tests/03.t tests/04.t tests/05.t tests/06.t \
       tests/37.t tests/38.t tests/39.t tests/40.t tests/41.t tests/42.t \
       tests/43.t
 TEST_EXTENSIONS=.t
-T_LOG_COMPILER=sh $(top_srcdir)/tests/test.sh
+T_LOG_COMPILER=$(top_srcdir)/tests/test.sh
 check_PROGRAMS=tests/pick-test
 tests_pick_test_SOURCES=$(TESTS) tests/test.sh tests/pick-test.c
 # Required by posix_openpt on Linux.
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,9 +16,9 @@ TESTS=tests/01.t tests/02.t tests/03.t tests/04.t tests/05.t tests/06.t \
       tests/37.t tests/38.t tests/39.t tests/40.t tests/41.t tests/42.t \
       tests/43.t
 TEST_EXTENSIONS=.t
-T_LOG_COMPILER=$(top_srcdir)/tests/test.sh
+T_LOG_COMPILER=tests/test.sh
 check_PROGRAMS=tests/pick-test
-tests_pick_test_SOURCES=$(TESTS) tests/test.sh tests/pick-test.c
+tests_pick_test_SOURCES=tests/pick-test.c
 # Required by posix_openpt on Linux.
 tests_pick_test_CFLAGS=-D_XOPEN_SOURCE=600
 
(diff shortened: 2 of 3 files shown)


calleerlandsson:
Great idea, @mptre. I don't seem to be allowed to flip that switch in Travis either. @mike-burns, are you?

mike-burns:
You should be able to do it now. Does it work?

(Also, it should be transferred to you now.)

calleerlandsson:
Thanks @mike-burns, I was able to active Travis now. I guess we should all update our git remotes now.

The build fails because of an old Autoconf version. In gitsh we dropped to `2.68` for exactly this reason: https://github.com/thoughtbot/gitsh/commit/dbc2b733ec5691023572c8ffb7ccdbab35223077

Given it works, I think doing so for pick makes sense as well.

mptre:
> Thanks @mike-burns, I was able to active Travis now. I guess we should all update our git remotes now.

We should notify package maintainers as well, don't know if all package
managers handle redirects.

> Given it works, I think doing so for pick makes sense as well.

Downgrading from `2.69` to `2.68` got us further but a different kind of
error occurred. Will examine the results more closely later today.

calleerlandsson:
> We should notify package maintainers as well, don't know if all package managers handle redirects.

Good point. I'll include this in my email about version 1.5.1.

calleerlandsson:
I wonder if the failures are related in any way to how we set the test suite up in `Makefile.am`. Seems to be failing on Homebrew's CI as well: https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/8789/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_pick/

mptre:
Tried to run a single test, bypassing make which also failed. I can't
access the log on Travis, prompted with the following error: `The
repository at calleerlandsson/pick was not found`

calleerlandsson:
Yeah, I'm seeing the same thing. I saw it earlier today too and then it worked for a bit. I'm suspecting Travis might be having issues.

I can try deactivating and re-activating later if it doesn't start working.

I'll be traveling for the rest of the week so unfortunately I'll be a bit unresponsive.

Thanks for working on this @mptre!

mptre:
Not the best first impression of this service. Lets be cautious from
here and evaluate if the invested time to get this working actually is
worth it.

calleerlandsson:
@mptre I can now access the output from Travis. Can you do so as well?

mptre:
Yes, solving the remaining problems on macOS probably requires talking
to Travis. I have to further ideas.

mptre:
Progress: works on macOS after making the TTY disconnect conditional.
Still some weird issues on Linux, though.

mptre:
I cherry-picked the relevant commit to get the tests running on macOS
under CI to the master branch.

@calleerlandsson: if you do a patch release we can hopefully solve #166,
Homebrew/homebrew-core#5079 and Homebrew/homebrew-core#5608.

calleerlandsson:
@mptre Thanks for doing that and for continuing trying to solve the issues on CI!

I've released [v1.5.2](https://github.com/calleerlandsson/pick/releases/tag/v1.5.2).

mptre:
Works at last! Using systrace made me realize that make tried to execute
each test-file rather than using the log compiler. Older versions of
automake probably interpreted the `TESTS` variable as a list of
executable commands. Moving the Trusty environment on Travis which uses
a later version of automake solves this issue.

This branch is a mess, I will remove it but make sure to keep the
relevant parts.

It seems like Travis is building each push twice since the PR-branch
belongs to the repository. Could this be prevented without disabling
testing of PRs?

calleerlandsson:
Great work, @mptre!

> It seems like Travis is building each push twice since the PR-branch
belongs to the repository. Could this be prevented without disabling
testing of PRs?

Confusingly enough, I actually think this is a feature of Travis and that it isn't building the same commit but rather the pushed commit and the automatically created merge commit that GitHub has already created in case you press the merge button on the website.

For example [the last build for this PR](https://travis-ci.org/calleerlandsson/pick/builds/167314275) builds this commit: https://github.com/calleerlandsson/pick/commit/cdfdcda2f6acda0e950c9fa1570377a9e9b8d864

Since we're doing fast-forward merges, I'm guessing these should always be the same? If so, maybe turning off the PR builds is the solution?

mptre:
Thanks for the info, lets keep the current duplicate build behavior for
now since it's a start.

A question regarding b3c5bb1, is the following addition essential to
creating a release? As I read it, `pick-test(1)` will be re-compiled if
any of the tests or test-runner change which seems redundant. The
dependency makes more sense the other way around but I might be missing
something here.

```diff
-tests_pick_test_SOURCES=tests/pick-test.c
+tests_pick_test_SOURCES=$(TESTS) tests/test.sh tests/pick-test.c
```

calleerlandsson:
The reason I made the change was to get `make distcheck` pass. `make distcheck` tries to do what automake calls [a `VPATH` build](https://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html#VPATH-Builds). It was failing because in this case, the test files weren't available. Adding them to the `_SOURCES` solved this. Maybe there's a better way of making them available?

mptre:
Thanks for the clarification. Merging now.

calleerlandsson:
Thanks!

PR #169, opened 2016-10-16
Simplify the project directory structure

calleerlandsson:
This started with an annoyance about the man page being kept alone two directories down for no apparent reason other than me putting it there from the start.

I then went on to move the source files from the `src` directory to the project root since they were so few.

Since source files were now kept in the project root I also tried moving the test runner source files to the root. Now, all files except the `README.md` in the `tests` directory serves the same purpose.

I don't have super strong opinions about this but rather thought this might be simpler. What are your thoughts, @mptre?


pick-test.c
```diff
@@ -42,7 +42,7 @@ main(int argc, char *argv[])
 	/* Ensure room for program and null terminator. */
 	if ((pickargv = calloc(argc + 2, sizeof(const char **))) == NULL)
 		err(1, NULL);
-	pickargv[0] = "pick";
+	pickargv[0] = "./pick";
```
mptre: Please keep the `PATH`-lookup.

test.sh
```diff
@@ -29,8 +29,7 @@ for testcase; do
     fi
   done <$testcase
 
-  env PATH="./src:${PATH}" tests/pick-test \
-    -k "$(printf "$keys")" -- $args <$stdin >$out 2>&1; e=$?
+  ./pick-test -k "$(printf "$keys")" -- $args <$stdin >$out 2>&1; e=$?
```
mptre: Please keep the `PATH`-lookup and change to:

```sh
env PATH=".:${PATH}"
```

mptre:
Love it, some nits commented inline.

calleerlandsson:
Thanks for the feedback, @mptre! I guess dropping 6e76063 when rebasing is the simplest way of reverting the changes you commented on. I'm curious why the `PATH` lookup is better, though? As mentioned in the message of 6e76063, my reasoning was that using relative paths removes any doubt regarding which program is being run. If I have pick installed on my `PATH`, and there is a bug or discrepancy in our Autotools setup, could the pick on my `PATH` be tested rather than the one created in the project?

mptre:
There have been times when being able to manipulate the PATH to change
the binary to run has been useful. But you're right, using relative paths
makes it more obvious so lets keep your change.

mptre:
After some thinking, I think getting rid of `src` and `man` is a great
start. But lets keep the tests as is for now.

calleerlandsson:
> But lets keep the tests as is for now.

Sounds OK to me. Just curious, what's the reason you'd like to keep the tests as is? Would you like to keep the change from PATH lookup to relative paths?

mptre:
> Sounds OK to me. Just curious, what's the reason you'd like to keep
> the tests as is?

No compelling argument yet. But I want to look into how other does it
before we make any changes.

> Would you like to keep the change from PATH lookup to relative paths?

Yes!

calleerlandsson:
Cool! I've updated my commits and will merge when the build is green.

tests/test.sh
```diff
@@ -29,8 +29,7 @@ for testcase; do
     fi
   done <$testcase
 
-  env PATH="./src:${PATH}" tests/pick-test \
-    -k "$(printf "$keys")" -- $args <$stdin >$out 2>&1; e=$?
+  tests/pick-test -k "$(printf "$keys")" -- $args <$stdin >$out 2>&1; e=$?
```
mptre: Change to `./tests/pick-test`?

calleerlandsson: I thought the two were identical?

calleerlandsson: I see. I'll merge this as is for now. Thanks for reviewing!

PR #172, opened 2016-10-20
Cleanup man page

mptre:
- Less verbose by reducing the usage of Nm

- Mention stdin/stdout for input/output

- Sort options, disregarding case

- Turn option descriptions into sentences

- Update cross-reference: ncurses -> curses

- Move keys into commands section, inspired by less(1)

- Simplify examples using Dl and add prefix with shell prompt, de facto
  standard to emphasis on the privileges needed to run the command

- Add diagnostics section

- Remove redundant quotes around authors and turn emails into hyperlinks

### DIFF
--- a/pick.1
+++ b/pick.1
@@ -11,81 +11,73 @@
 .Op Fl x | Fl X
 .Op Fl q Ar query
 .Sh DESCRIPTION
+The
 .Nm
-allows users to select from a set of choices using an
-.Xr ncurses 3X
+utility allows users to select from a set of choices using an
+.Xr curses 3
 interface with fuzzy search functionality.
 .Pp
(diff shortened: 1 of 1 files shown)


pick.1
```diff
@@ -11,81 +11,73 @@
 .Op Fl x | Fl X
 .Op Fl q Ar query
 .Sh DESCRIPTION
+The
 .Nm
-allows users to select from a set of choices using an
-.Xr ncurses 3X
+utility allows users to select from a set of choices using an
```
calleerlandsson: Should it be "a" instead of "an"?

mptre:
Thanks, merged.

PR #173, opened 2016-10-24
Remove unused autoconf checks

mptre:
Validating existence of things that hasn't changed since the 80s seems
wasteful. Especially when the types, headers and functions to look for
is out-of-date in respect to what's used by pick.

Looking for the two non-portable functions pledge and reallocarray makes
sense and is therefore kept as is.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -1,60 +1,13 @@
 /* config.h.in.  Generated from configure.ac by autoheader.  */
 
-/* Define to 1 if you have the <inttypes.h> header file. */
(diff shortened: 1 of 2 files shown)


calleerlandsson:
I agree. Thanks for figuring this out!

Runs and builds on my Mac. I'll try it on my Debian laptop tonight.

calleerlandsson:
Actually, since it builds on Travis it should just work. Merge at will!

mptre:
Thanks, merged.

PR #174, opened 2016-10-25
Fix underline spill over bug

mptre:
If the end of the match is greater than the number of columns the
underline attribute will spill over on the next line since it's never
exited. Running the following command before/after applying this commit
highlights the bug:

```sh
for _ in 0 1; do
  jot -n -s '' $(($COLUMNS - 1)) 0 0
  echo ab
done | pick -q ab
```


calleerlandsson:
Awesome! Thanks for finding that one. Confirmed as working on Debian 8. Merge at will.

mptre:
Thanks, merged.

PR #176, opened 2016-10-30
Don't normalize score with length of choice

mptre:
The purpose of this PR is to lay the foundation for a discussion. I
believe a discussion always should start with a patch.

Why is the score normalized with the length of the choice? This
implies that shorter choices should be favored.

In the following scenario, I would assume `abc, a` would be the order
of the choices with the query `a`:

``` sh
printf 'abc\na\n' | pick -q a
```

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -547,7 +547,7 @@ score(struct choice *choice)
 	}
 
 	match_length = choice->match_end - choice->match_start;
-	choice->score = (float)query_length / match_length / choice->length;
+	choice->score = (float)query_length/match_length;
 }
 
 size_t
--- a/tests/08.t
+++ b/tests/08.t
@@ -6,4 +6,4 @@ ABC
 AB
 A
 stdout:
-A
+ABC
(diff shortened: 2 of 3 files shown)


calleerlandsson:
Thanks for thinking about this, @mptre. I'm sorry for not getting back to you earlier. I agree that a patch is a great start for a discussion.

The score is currently normalized by the choice length in order to favor more complete matches. This is how I comprehend the code:

The length of the query is divided by the length of the shortest match. This produces a number between 0 and 1 which represents how exact the match is. This is further divided by the total length of the choice, weighing in how similar the query to the choice.

A practical example of this is favoring `app/models/user.rb` over `test/models/user_test.rb` when searching for `user`.

In your example I would assume the opposite order since the query `a` and the choice `a` is a 100% match.

Does that make sense?

mptre:
Thanks for the explanation, makes sense. I guess it's a matter of
preference then. I often want to find a shorter substring in something
potentially longer. The length of the match compared to the query is
then a better relevance indicator because the suffix and prefix is not
relevant IMHO. But I like the fact that the current implementation has
reasoning behind its behavior. Let's keep it that way.

PR #178, opened 2016-12-07
Recognize carriage return as enter

### LINKED ISSUE #177: Pick not working with Bash's `shell-expand-line` binding
The option `shell-expand-line` in bash expands the line with a subshell present into the result of that subshell.

Example:

```sh
% echo $(printf "%f" 3.5)
# -----------------------^
# Enter keypress: \e\C-e
% echo 3.500000
# -------------^
```

The use-case for `pick(1)` would be to perform an action on a file that was selected from pick itself.

i.e

```sh
% cat $(git ls-files | pick)
# --------------------------^
# Enter keypress: \e\C-e
```

The problem is that neither `| pick` or `| pick -X` can perform an enter key-press to (in lack of a better word) "complete" the subshell. This has been tested with `"$()"` aswell as double back-tricks.
Strangely though, `C-j` works in place for the `CR`.

This works in `zsh` using their `expand-or-complete` command (mapped to `C-i`) with both `CR` and `C-j` to "complete" the subshell.

mptre:
Fixes #177.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -13,7 +13,7 @@ TESTS=tests/01.t tests/02.t tests/03.t tests/04.t tests/05.t tests/06.t \
       tests/25.t tests/26.t tests/27.t tests/28.t tests/29.t tests/30.t \
       tests/31.t tests/32.t tests/33.t tests/34.t tests/35.t tests/36.t \
       tests/37.t tests/38.t tests/39.t tests/40.t tests/41.t tests/42.t \
-      tests/43.t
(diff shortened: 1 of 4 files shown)


teoljungberg:
Looks good to me! Also, thank you for adding tests for the feature.

mptre:
Great! @calleerlandsson time for another release?

calleerlandsson:
Thanks! Yes, I'll create one!

PR #179, opened 2016-12-23
Inline score function

mptre:
This is a hot path and it looks more like a left-over from a previous
design.


calleerlandsson:
Nice! Looks good to merge!

mptre:
Thanks!

PR #180, opened 2016-12-27
Only define __dead if it's missing

mptre:
I propose we only define `__dead` to the empty string if it's missing
since the current `ifdef`-logic is rather complex and we don't want to
maintain support for all various platforms and compilers.

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -5,19 +5,9 @@
 #include "config.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


calleerlandsson:
Good call. Looks good to merge!

mptre:
Thanks, merged.

PR #181, opened 2017-01-17
Setup crucial environment for pick in pick-test

mptre:
Instead of doing [this][homebrew], `pick-test(1)` should setup the environment
needed by `pick(1)`.
Once the released the referenced line could be removed from the homebrew
recipe.

Thoughts @calleerlandsson and @teoljungberg?

[homebrew]: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pick.rb#L15

### DIFF
--- a/tests/pick-test.c
+++ b/tests/pick-test.c
@@ -18,8 +18,19 @@ static char	*parsekeys(const char *);
 static void	 sighandler(int);
 static void	 usage(void);
 
-static char	**pickargv;
-static int	  gotsig;
+static char		**pickargv;
+/*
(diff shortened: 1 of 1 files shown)


calleerlandsson:
This seems like a great idea and will probably make the test suite more reliable across different systems!

mptre:
Thanks, merged. @calleerlandsson would you mind crafting a release?

teoljungberg:
I'll update Homebrew with the new release as soon as its cut.

teoljungberg:
Submitted to Homebrew as: https://github.com/Homebrew/homebrew-core/pull/9246

PR #183, opened 2017-04-11
Remove usage of sysexits.h

mptre:
A recent discussion on the OpenBSD [tech-mailing-list] brought to my
attention that sysexits.h originates from sendmail(8) and further usage
is not encouraged.

Ping @calleerlandsson @mike-burns 

[tech-mailing-list]: http://marc.info/?l=openbsd-tech&m=149191614721371&w=2

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -11,7 +11,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sysexits.h>
 #include <termios.h>
 #include <unistd.h>
 #include <wchar.h>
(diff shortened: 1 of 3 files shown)


calleerlandsson:
I think `sysexits.h` is an interesting idea but since it never really caught on, it's not particularly useful.

Looks good to me!

mptre:
Thanks, merged!

@calleerlandsson care to craft a release? I'm the process of making pick
available on OpenBSD and just committed a few fixes. I would argue that
this release require a minor bump due to the change of exit codes.

calleerlandsson:
Sure thing! Should we get #185 merged in before?

mptre:
That would be nice. I would really appreciate if you could take a closer
look at first.

PR #185, opened 2017-04-16
Calculate the width of each displayed character

### LINKED ISSUE #184: Texts are not aligned and position change while I scroll down...
work fine while all texts are English or numbers...
2017-04-15-165730_1919x1020_scrot
2017-04-15-170349_1919x1019_scrot

mptre:
In an attempt to fix #184 caused by lines including Unicode characters
that occupy more than one column.

While at it, fix another bug related to the column counter when a line
contains more than one escape sequence.

Reviews and testing would be much appreciated I haven't discovered any
regression nor performance decrease running this diff so far.

/cc @calleerlandsson @mike-burns @teoljungberg.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,7 @@
 AUTOMAKE_OPTIONS=foreign
 
 AM_CFLAGS=-pedantic -Wall -Werror -Wextra
+AM_CPPFLAGS=-D_GNU_SOURCE
 
 bin_PROGRAMS=pick
 dist_pick_SOURCES=pick.c compat.h compat.c
@@ -19,8 +20,6 @@ T_LOG_COMPILER=$(top_srcdir)/tests/pick-test.sh
 check_PROGRAMS=tests/pick-test
 tests_pick_test_SOURCES=tests/pick-test.c
 tests_pick_test_CFLAGS=$(AM_CFLAGS)
-# Required by posix_openpt on Linux.
-tests_pick_test_CFLAGS+=-D_XOPEN_SOURCE=600
 
 EXTRA_DIST=INSTALL.md INSTALL.md.in LICENSE README.md tests/pick-test.sh $(TESTS)
 
--- a/pick.c
+++ b/pick.c
@@ -667,65 +667,95 @@ tty_restore(void)
 }
 
 void
-print_line(const char *string, size_t length, int so, ssize_t ulso, ssize_t uleo)
+print_line(const char *str, size_t len, int standout,
+    ssize_t enter_underline, ssize_t exit_underline)
 {
 	size_t	i;
-	int	c, col, tabwidth;
-	int	in_esc_seq = 0;
-	int	non_printable = 0;
+	wchar_t	wc;
+	int	col, in_esc_seq, nbytes, width;
 
-	if (so)
+	if (standout)
 		tty_putp(enter_standout_mode, 1);
 
-	for (col = i = 0; i < length && col < columns; i++) {
-		if (i == (size_t)ulso)
+	col = i = in_esc_seq = 0;
+	while (col < columns) {
+		if (enter_underline == (ssize_t)i)
 			tty_putp(enter_underline_mode, 1);
+		else if (exit_underline == (ssize_t)i)
+			tty_putp(exit_underline_mode, 1);
+		if (i == len)
+			break;
 
-		/*
-		 * Count the number of outputted ANSI escape sequences
-		 * in order to adjust the column count since they do not
-		 * occupy any screen real-estate.
-		 */
-		if (in_esc_seq) {
-			non_printable++;
-			if (string[i] >= '@' && string[i] <= '~')
-				in_esc_seq = 0;
-		} else if (i > 0 && string[i - 1] == '\033'
-		    && string[i] == '[') {
-			in_esc_seq = 1;
-			non_printable = 2;
-		}
-
-		c = string[i];
-		if (c == '\t') {
-			/* Ceil column count to multiple of 8. */
-			col += tabwidth = 8 - (col & 7);
-			while (tabwidth-- > 0)
+		if (str[i] == '\t') {
+			width = 8 - (col & 7);	/* ceil to multiple of 8 */
+			if (col + width > columns)
+				break;
+			for (; width > 0; width--)
 				if (tty_putc(' ') == ERR)
 					err(1, "tty_putc");
-		} else {
-			/*
-			 * A null character will be present prior the
-			 * terminating null character if descriptions is
-			 * enabled.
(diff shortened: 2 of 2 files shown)


pick.c
```diff
@@ -667,19 +667,23 @@ tty_restore(void)
 }
 
 void
-print_line(const char *string, size_t length, int so, ssize_t ulso, ssize_t uleo)
+print_line(const char *str, size_t len, int so, ssize_t ulso, ssize_t uleo)
```
teoljungberg: I, for one, am of the view to write out as many acronyms as possible. But since we have `ssize_t` unt so weiter already, I'd say this falls in line with that style better.

mptre:
> Is there a test we can add for this change? Or is the testing setup we have now non-applicable for a change such as this?

Unfortunately no. Since this change only concerns the displayed
TTY-output, it's not testable.

> I, for one, am of the view to write out as many acronyms as possible. But since we have `ssize_t` unt so weiter already, I'd say this falls in line with that style better.

Agree, just renamed the arguments. Thanks!

configure.ac
```diff
@@ -3,6 +3,14 @@ AC_INIT([pick], [1.5.4], [pick-maintainers@calleerlandsson.com])
 AM_INIT_AUTOMAKE([subdir-objects])
 AC_CONFIG_HEADERS([config.h])
 AC_PROG_CC
+
+AC_CANONICAL_HOST
+build_linux=no
+case "${host_os}" in
+linux*) build_linux=yes ;;
+esac
+AM_CONDITIONAL([LINUX], [test "$build_linux" = "yes"])
```
calleerlandsson: Is `$build_linux` used solely to define `LINUX` so that it can be picked up in `Makefile.am` or is it used elsewhere?

configure.ac
```diff
@@ -3,6 +3,14 @@ AC_INIT([pick], [1.5.4], [pick-maintainers@calleerlandsson.com])
 AM_INIT_AUTOMAKE([subdir-objects])
 AC_CONFIG_HEADERS([config.h])
 AC_PROG_CC
+
+AC_CANONICAL_HOST
+build_linux=no
+case "${host_os}" in
+linux*) build_linux=yes ;;
```
calleerlandsson: I haven't used `AC_CANONICAL_HOST` before and didn't find any documentation detailing the formatting of `host_os`. Is `linux*` a reliable pattern for catching all Linux-based OSs?

Makefile.am
```diff
@@ -2,6 +2,10 @@ AUTOMAKE_OPTIONS=foreign
 
 AM_CFLAGS=-pedantic -Wall -Werror -Wextra
 
+if LINUX
```
calleerlandsson: Could defining `_XOPEN_SOURCE` on non-linux platforms cause problems? If not, what do you think of always defining it? That way we could avoid the complexities of testing for the host os.

pick.c
```diff
@@ -667,65 +667,95 @@ tty_restore(void)
...
 	}
-	for (col -= non_printable; col < columns; col++)
+	for (; col < columns; col++)
 		if (tty_putc(' ') == EOF)
 			err(1, "tty_putc");
 
-	/* If uleo is greater than columns the underline attribute will spill
-	 * over on the next line unless all attributes are exited. */
+	/*
+	 * If uleo is greater than columns the underline attribute will spill
```
calleerlandsson: Should this comment say `exit_underline` instead of `uleo` that the parameter names have been changed?

pick.c
```diff
@@ -667,65 +667,95 @@ tty_restore(void)
...
-			while (tabwidth-- > 0)
-				if (tty_putc(' ') == ERR)
-					err(1, "tty_putc");
-		} else {
+		/*
+		 * Due to the explicit NUL-check above, the case where
+		 * mbtowc(3) returns 0 is not handled here.
+		 */
+		if ((nbytes = mbtowc(&wc, &str[i], MB_CUR_MAX)) == -1) {
+			mbtowc(NULL, NULL, MB_CUR_MAX);
```
calleerlandsson: Is this for resetting internal state of `mbtowc`?

Makefile.am
mptre: Replying to all comments above here. Unfortunately no, defining it on
*BSD makes BSD-specific functions like `pledge` and `reallocarray`
unavailable.

Maybe @mike-burns could shed some insight on how-to solve this?

The OS detection is taken from here [1] [2].

[1]: https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Canonicalizing.html
[2]: http://www.clearchain.com/blog/posts/autotools

pick.c
mptre: Correct

pick.c
mptre: Nice catch, fixed!

mptre:
> I did find a regression but haven't had time to figure out exactly why it's happening. Steps to reproduce:

Can't reproduce on OpenBSD but I will try it on macOS this evening.

Makefile.am
calleerlandsson: I see. Thanks for the links! I'm happy to try this and correct course if we run into issues with some distro.

calleerlandsson:
Thanks @mptre. Let me know if I can help you in any way!

Makefile.am
mike-burns: I don't see why, from a quick skim of the repo, `pledge` disappears when defining `_XOPEN_SOURCE`. But I do see this in glibc: https://github.com/bminor/glibc/blob/963394a22b38c4ec92b6875a6c06d3b15d5c0d21/include/features.h#L186-L208

So maybe: skip the OS check (which should be a glibc check ... somehow?) and instead do:

```
AM_CPPFLAGS+=-D_GNU_SOURCE
```

mptre: Works, thanks!

mptre:
Turns out the passing that specific emoji to `wcwidth(3)` returns 0 on
macOS. This is fortunately not true for all emojis. I don't think we
should bother handling this edge-case.

OK to merge?

calleerlandsson:
Interesting. Searching for fixes, I found a lot of tickets and issues discussing this problem and a lot of homegrown solutions special casing certain characters. I'm of the opinion that pick(1) should solve these problems.

Looks good to merge.

mptre:
> Interesting. Searching for fixes, I found a lot of tickets and issues discussing this problem and a lot of homegrown solutions special casing certain characters. I'm of the opinion that pick(1) should solve these problems.

Can you create a separate issue and share your findings there?

> Looks good to merge.

Done! Thanks for the feedback. Time to release?

calleerlandsson:
Sure thing!

PR #187, opened 2017-04-20
Fix tab output bug

mptre:
Increment the column counter prior decrementing the calculated width of
the tab.

Regression introduced in commit 369350b (Calculate the width of each
displayed character).

Sorry, can we do a patch release @calleerlandsson?


calleerlandsson:
Sure thing! Merge and update the changelog when you're ready and I'll do a patch release.

mptre:
Done!

PR #192, opened 2017-06-17
Map CR to NL

mptre:
Instead of explicitly list all key bindings that could be either CR or
NL, make use of the ICRNL input mode.

@teoljungberg could you verify that this PR work as intented on your setup?


calleerlandsson:
Works well on my system, if it works well for @teoljungberg I think this is good to merge!

teoljungberg:
No regressions were found, looks good to merge for me.

mptre:
Thanks, merged!

PR #194, opened 2017-06-18
Add Keys for Home/End and Ctrl-H for Backspace

DBOTW:


pick.c
```diff
@@ -31,23 +31,25 @@
 } while (0)
 
 enum key {
-	UNKNOWN,
```
mptre: Although, I do agree this list should be sorted. But lets skip that for
now and produce the smallest possible diff. Just add `END` and `HOME`
before `PRINTABLE`.

tests/47.t
```diff
@@ -0,0 +1,9 @@
+description: select the first choice
+keys: \033OH \\n # HOME ENTER
```
mptre: Could we move to the second choice prior pressing `HOME`, just to make
sure it actually works:

```
keys: \016 \033OH \\n # DOWN HOME ENTER
```

tests/49.t
```diff
@@ -0,0 +1,7 @@
+description: backspace
+keys: ab \b c \\n ## BACKSPACE ENTER
```
mptre: You could incorporate this in `tests/03.t` instead of adding a new test:

```
keys: abb \177 \b c \\n ## BACKSPACE BACKSPACE ENTER
```

Don't forget to remove `tests/49.t` from `Makefile.am`.

mptre:
Great work! Some minor inline comments, other than that this looks good.
You don't need to create a new PR once you fixed the issues, just commit
a [fixup](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html), rebase then push.

@calleerlandsson could you also review this?

calleerlandsson:
Hey @DBOTW and @mptre, sorry I haven't had time to look at this PR until now. I think this sounds like a good addition and would like to merge it in. Would you like to re-open the PR or open a new one, @DBOTW?

PR #197, opened 2017-06-24
Terminfo

mptre:
Read keys defined as terminal capabilities from terminfo

A first step towards making pick truly portable. Defining a key using
the new CAP() macro will read the corresponding value for the given
capability using tigetstr(), the result will be stored in the keys
structure in order to only invoke tigetstr() once per capability.

@DBOTW could you verify that this PR works as intended on your machine?


DBOTW:
Tested and works flawlessly, so other KEY() macros could be replaced with CAP() too ...

mptre:
@DBOTW great! I would like to do a release before we migrate the rest of
the keys just make any regression easier to bisect. Feel free to open a
new PR with your `KEY -> CAP` changes.

/cc @calleerlandsson

pick.c
```diff
@@ -838,26 +836,29 @@ get_key(char *buf, size_t size, size_t *nread)
...
 	for (; size > 0; size--) {
 		buf[(*nread)++] = tty_getc();
 
-		for (i = 0; keys[i].s != NULL; i++) {
-			if (*nread > keys[i].length
-			    || strncmp(buf, keys[i].s, *nread) != 0)
+		for (i = 0; keys[i].key != UNKNOWN; i++) {
+			if (keys[i].str == NULL)
+				keys[i].str = tty_getcap(keys[i].cap);
+			len = strlen(keys[i].str);
```
calleerlandsson: What do you think about caching this in `keys` like we do with the string?

pick.c
```diff
@@ -800,11 +801,12 @@ print_choices(int offset, int selection)
...
 get_key(char *buf, size_t size, size_t *nread)
 {
-#define	KEY(k, s)	{ k, s, sizeof(s) - 1 }
+#define	CAP(k, s)	{ k,	s,	NULL }
+#define	KEY(k, s)	{ k,	NULL,	s }
 	static struct {
 		enum key	 key;
-		const char	*s;
-		size_t		 length;
+		char		*cap;
```
calleerlandsson: I was wondering why this wasn't `const char` but then I saw `tigetstr` takes a `char *`. I'm not quite sure why though...

mptre: Yeah, that's exactly why it can't be `const`. I don't have an
explanation either.

mptre:
Great feedback! The length is now cached.

Issue #196, opened 2017-06-23
Unexpected behavior on certain inputs


EpicDavi:
I recently started playing around with pick (1.6.1) and it's fantastic! I discovered some weird behavior while messing around with it.

Here is an example: `pick -h | pick`. Running this initially results in a blank screen but as characters are inputted, the current string is duplicated until it reaches the top of the screen.

Piping in `pick -h` is probably not a typical use case, but the behavior that results is strange.

EDIT: I am running macOS 10.12.5, if this helps at all.

EpicDavi:
This might be related to issue #62 ?

calleerlandsson:
Hey @EpicDavi, thanks for trying out pick. I'm happy you like it!

When running `pick -h | pick` from the shell, the `stdout` stream of `pick -h` will be connected to the `stdin` stream of `pick`. Since `pick -h` outputs the usage message to `stderr` rather than to `stdout`, this passes no results to `pick`. That should explain the black screen you see.

To exit pick you can press Ctrl-C. When you do, you should see the usage message that `pick -h` printed to your terminal (which was attached to its `stderr`). You can also type something while pick is running and press Alt-Enter to have pick output that when exiting.

> as characters are inputted, the current string is duplicated until it reaches the top of the screen.

Could you elaborate on this? Maybe upload a screen recording? This does sound like a bug but I can't reproduce it on macOS 10.12.5 running pick from bash in Terminal. Are you running a different shell or terminal emulator?

EpicDavi:
Just tried it on another machine (macOs 10.12) and am having the same issue.

However, I tried it on the built-in Terminal (instead of iTerm2) and the issue no longer occurs. This makes me think it is an issue with iTerm2 (3.0.15)?

Anyways, here is what the issue looks like, if you are curious: https://asciinema.org/a/xqJPG5vAVHkkIvNdblbiOuKDJ

mptre:
Thanks for the report, I'm able to reproduce this using `xterm(1)`:

```sh
$ pick </dev/null
```

Please try out #198 and see if it solves the problem.

mptre:
Fix has been merged, closing for now.

PR #198, opened 2017-06-25
Don't move the cursor upwards if no choices where printed

### LINKED ISSUE #98: pick hangs when piped empty results
When piping unknown results to pick, e.g.: 

ag -l 'foo.*bar' | pick 

an empty search result appears to hang pick. Crl-c is the only escape; alternatively, creating a bash function like: 

pipe_if_not_empty () {
  input=$(cat; echo a);
  if [ "x$input" != x"a" ]; then
    { printf %s "${input%a}"; } | "$@"
  fi
}
then one can resort to a solution like:
 
ag -l 'foo.*bar' | pipe_if_not_empty pick 

However, pick should handle this itself.

mptre:
In an attempt to fix #98 and #196.

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -312,7 +312,11 @@ selected_choice(void)
 				err(1, "tty_putc");
 			tty_putp(clr_eos, 1);
 			tty_putp(tparm(parm_up_cursor, choices_count + 1), 1);
-		} else {
+		} else if (choices_count > 0) {
+			/*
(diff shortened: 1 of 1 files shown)


calleerlandsson:
Good catch, @mptre, and thanks for adding the comment explaining why this happens. I think the code looks good to merge!

@EpicDavi, does this PR solve your issues in iTerm2?

EpicDavi:
Yes, this fixes the issue for me!

Issue #199, opened 2017-06-25
selection bar vanishes on "Page-Down"


DBOTW:
`% seq 1 $LINES | pick`  and press "Down" (once or serveral times) then "Page-Down": the selection bar vanishes. The following should fix this:

```diff
--- pick.c.orig	2017-06-25 23:59:43.257283180 +0200
+++ pick.c	2017-06-26 00:32:17.077325615 +0200
@@ -449,8 +449,11 @@
 		case PAGE_DOWN:
 			if (selection + choices_lines < choices_count)
 				yscroll = selection += choices_lines;
-			else
+			else {
 				selection = choices_count - 1;
+				if (selection - yscroll >= choices_lines)
+					yscroll = choices_count - choices_lines;
+			}
 			break;
 		case PAGE_UP:
 			if (selection - choices_lines > 0)
```

mptre:
Good catch! Care to craft a PR? Also, make sure to wrap the `if` in
braces as well for consistency.

mptre:
Fix has been merged, closing for now. Thanks!

PR #200, opened 2017-06-26
Fix to #199: selection bar vanishes on Page-Down

DBOTW:


mptre:
Thanks, merged as 92e492b.

PR #201, opened 2017-06-28
Some more Emacs key bindings

DBOTW:
Like "Ctrl-P/Ctrl-N" for "Up/Down" there should be some key bindings for other "special keys" as well.


mptre:
Are those key bindings de facto standard for moving vertically
somewhere?

DBOTW:
"less"uses them too ... may be useful on machines where "special keys" are not available; I think that was the reason for adding "Ctrl-P" and "Ctrl-N" ...

calleerlandsson:
I'm happy to support key bindings used by common utilities and I think this looks good to merge! Thanks for working on this @DBOTW!

mptre:
That went a bit too fast, I would like to see tests for the new keys.

calleerlandsson:
Oh yeah, good point, @mptre. That way we won't break the key bindings going forward.

PR #203, opened 2017-07-01
Add -Q switch: output selection with double quotes

DBOTW:
for preserving "white spaces" in the selected choice ...

### DIFF
--- a/pick.1
+++ b/pick.1
@@ -6,7 +6,7 @@
 .Nd fuzzy select anything
 .Sh SYNOPSIS
 .Nm
-.Op Fl hvS
+.Op Fl hvQS
 .Op Fl d Op Fl o
 .Op Fl x | Fl X
 .Op Fl q Ar query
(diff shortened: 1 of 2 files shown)


mptre:
Hi Jenz, pick is intended to ease the selection of a singe line from
many. In order to keep things simple it doesn't alter its input, with
the exception of the `-d` and `-o` options. For the problem you're
trying to solve I would advocate to use a better tool for the job:

```sh
$ echo quote | pick | xargs printf '"%s"\n'
"quote"
```

There's also the scenario when the line already contains one or many
quotes:

```sh
$ echo quot\"e | pick -Q
"quot"e"
```

... which instead could be solved using for instance `sed(1)`:

```sh
$ echo quot\"e | ./pick | sed -n -e 's/"/\\&/g' -e 's/.*/"&"/p'
"quot\"e"
```

DBOTW:
Hi Anton, you're right: further processing should not be done by `pick` ...

mptre:
@Jenz btw, please go ahead and open a PR with your migrations from `KEY`
to `CAP`.

PR #206, opened 2017-07-03
Use new CAP() macro for arrow keys

DBOTW:


mptre:
Great work!

Regarding the build failure on macOS: turns out `kbs=^H` for xterm on
macOS. I would therefore only use `\b` for backspace in the tests in
order to make all tests pass on all platforms.

DBOTW:
Thanks again @mptre !

mptre:
With this diff applied I can't use the arrow keys on neither OpenBSD nor
macOS. As an example, pressing `DOWN` emits `\e[B` while the terminfo
entry equals `\eOB`. Tested with Terminal.app and xterm with and without
tmux.

Is any else seeing this? /cc @calleerlandsson @teoljungberg

teoljungberg:
My arrow keys broke aswell. Tested on macOS, in Terminal.app, in tmux and outside.

calleerlandsson:
Same here in bash in Terminal.app on macOS. Backspace also doesn't work.

Do we need to anything other than calling `setupterm` the way we do, before we call `tigetstr`?

DBOTW:
Yes, we do: `tty_putp(keypad_xmit, 0);`  after `setupterm` will fix this issue with "xterm" ...

DBOTW:
Tested with/without `dvtm`, `tmux` and `screen` in `xterm` and `terminator` ... I use `rxvt-unicode` here, where this issue did not occur ... more testings will follow if needed ...

pick.c
```diff
@@ -649,6 +649,7 @@ tty_init(void)
 		err(1, "fopen");
 
 	setupterm((char *)0, fileno(tty_out), (int *)0);
+	tty_putp(keypad_xmit, 0);
```
mptre: Nice find! I would prefer grouping the `tty_putp` calls and make sure to
exit from the same mode.

```diff
diff --git a/pick.c b/pick.c
index 2c23aeb..df185b9 100644
--- a/pick.c
+++ b/pick.c
@@ -655,6 +655,8 @@ tty_init(void)
 	if (use_alternate_screen)
 		tty_putp(enter_ca_mode, 0);
 
+	tty_putp(keypad_xmit, 1);
+
 	signal(SIGINT, handle_sigint);
 }
 
@@ -677,6 +679,7 @@ tty_restore(void)
 	tcsetattr(fileno(tty_in), TCSANOW, &original_attributes);
 	fclose(tty_in);
 
+	tty_putp(keypad_local, 1);
 	tty_putp(carriage_return, 1);	/* move cursor to first column */
 	tty_putp(clr_eos, 1);
 
```

mptre:
Everything works except backspace on macOS, I think we should keep the
following:

```c
KEY(BACKSPACE,  "\177");
```

I appreciate the effort on adding tests for the Emacs bindings, could
that be a separate PR?

DBOTW:
What the hell is wrong with macOS? ;-)

mptre:
Check out the [cap] branch, it includes some cleanups and is what I
propose on committing.

[cap]: https://github.com/calleerlandsson/pick/tree/cap

DBOTW:
Tested and looks good!

mptre:
Great! Lets wait for another confirmation, I will then go ahead and merge this.

/cc @calleerlandsson @teoljungberg

calleerlandsson:
Hey @mptre and @DBOTW, I'm sorry I've been unresponsive; I've been out traveling the last week. The cap branch works well on my machine and looks great to merge! Thanks for your hard work!

mptre:
Thanks, merged!

Issue #115, opened 2016-05-03
Terminal resize while pick is running


prahlad37:
If we resize the terminal after pick is initiated, the choices doesn't print as expected.
This may be because the variables 'lines', 'columns' are not getting updated, if the terminal size is disturbed after 'setupterm' been executed.
One possible solution may be to capture the event SIGWINCH and update the variables accordingly.

mptre:
On Tue, May 03, 2016 at 06:06:12AM -0700, prahlad wrote:
> If we resize the terminal after pick is initiated, the choices doesn't print as expected.
> This may be because the variables 'lines', 'columns' are not getting updated, if the terminal size is disturbed after 'setupterm' been executed.
> One possible solution may be to capture the event SIGWINCH and update the variables accordingly.

Agreed, even tho it is not a concern of mine but a interactive program
should behave well and resize accordingly. Some thoughts regarding the
implementation:

- We don't want to be interrupted while doing important work. The
  SIGWINCH signal should therefore only be allowed while the program is
  waiting for user-input (inside the get_key function). This can be
  achieved using ppoll(2).

- The signal handler should be kept as short as possible to avoid any
  potential data-race:

    void
    onsignal(int sig)
    {
      resize = sig == SIGWINCH;
    }

- The get_key function could return -1 if it was interrupted which then
  can be checked in the selected_choice function and if the global
  variable resize is set call ioctl in order to obtain the new
  dimensions and lastly reset the resize variable.

- The SIGWINCH signal is not portable(?) and its presence should be
  detected using autoconf.

Are you interested in implementing this?

prahlad37:
If I understand this correctly, you want to allow the signal only while we are waiting and block during processing part. In that case there is a chance that we may miss the event if we signal get blocked.
May be signalfd will suit here, which will block the signal and sends it through a fd and the handling can be defered through poll.

mptre:
On Wed, May 04, 2016 at 05:05:05AM -0700, prahlad wrote:
> If I understand this correctly, you want to allow the signal only while we are waiting and block during processing part. In that case there is a chance that we may miss the event if we signal get blocked.
> May be signalfd will suit here, which will block the signal and sends it through a fd and the handling can be defered through poll.

Correct, but I still think it's a reasonable trade-off.

mptre:
Resizing support has been implemented and will be released soon.

PR #209, opened 2017-07-08
Add redraw-function with Ctrl-L

DBOTW:
Instead of checking the position of the "selection bar" and setting "yscroll" accordingly, both are set to zero: if the display is already messed up by resizing the terminal window, it's IMHO better to start over ... ;-)

### DIFF
--- a/pick.1
+++ b/pick.1
@@ -78,6 +78,8 @@ input field.
 Delete to the beginning of the line in the search query input field.
 .It Ic Ctrl-K
 Delete to the end of the line in the search query input field.
+.It Ic Ctrl-L
+Redraw all choices and move the selection to the first choice.
 .It Ic Printable characters
 Added to the search query and will refine the current search.
 .El
--- a/pick.c
+++ b/pick.c
@@ -39,6 +39,7 @@ enum key {
 	CTRL_A,
 	CTRL_E,
 	CTRL_K,
+	CTRL_L,
 	CTRL_U,
 	CTRL_W,
 	UP,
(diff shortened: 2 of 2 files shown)


mptre:
Interesting approach. However, I would prefer to see my proposal on how
to handle `SIGWINCH` in #115 being implemented. With that logic in
place supporting `^L` sure makes sense.

DBOTW:
Thanks, I'm already "testing" a version of `redraw` that handle `SIGWINCH` and follows your proposals in https://github.com/calleerlandsson/pick/issues/115#issuecomment-216641683 but not yet satisfied with the result (or even the code). I don't want to use `ppoll`: it's not in `POSIX` (neither is `SIGWINCH`) and needs a newer `glibc`. So this resembles what the shell does on `Ctrl-L`:  clear the screen, put the current line on the top and start over. Only a few lines of code and you can still use `pick` even if the terminal window was resized and only "garbage" is left on the screen ...

mptre:
Valid points! Then I propose the following:

- Add two global variables:

  ```c
  static int tty_columns, tty_lines;
  ```

- Add a new function `tty_size` which does something like:

  ```c
  void tty_size(void)
  {
    struct winsize ws;

    if (ioctl(fileno(tty_in), TIOCGWINSZ, &ws) != -1) {
      tty_columns = ws.ws_col;
      tty_lines = ws.ws_row;
    } else {
      tty_columns = 80;
      tty_lines = 24;
    }
  }
  ```

- Call `tty_size` from `tty_init` at a appropriate place.

- Replace usage of `columns` and `lines` with the new global variables.

- Handle `CTRL_L` in `selected_choice`, should just be a matter of
  calling `tty_size` and reset some variables such as the selection etc.

- Add a test case, you could make sure that the selection is reset by
  first moving it downwards, pressing `^L` then enter, the first choice
  should then be selected.

Caution: code above not tested.

DBOTW:
What is the advantage over just calling `restartterm`?

mptre:
At some point I want to handle `SIGWINCH`, this is a first step in that
direction.

mptre:
Also, `tty_size` should set `choices_lines` as well.

DBOTW:
OK, so please checkout https://github.com/DBOTW/pick/tree/ttysize

mptre:
Nice! Please open a PR, got some feedback.

PR #210, opened 2017-07-09
Declare handle_sigint as dead

DBOTW:
Not really necessary, but `usage` is also declared as dead ...

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -67,7 +67,7 @@ static char			*eager_strpbrk(const char *, const char *);
 static void			 filter_choices(void);
 static char			*get_choices(void);
 static enum key			 get_key(char *, size_t, size_t *);
-static void			 handle_sigint(int);
+__dead static void		 handle_sigint(int);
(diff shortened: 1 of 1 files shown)


calleerlandsson:
This looks good to me! @mptre, do you have any thoughts?

DBOTW:
Meanwhile I've changed my mind: it makes no difference in the resulting binary, because the compiler "sees" the `exit`-instruction. So better remove `__dead` from `usage` too and "help" the compiler only when warnings about this appear ...

mptre:
I think we should leave both `handle_sigint` and `usage` as is for now.
OK to close?

DBOTW:
It's only about removing the `__dead` attribute from `usage` ... I think it confuses: why is `usage` marked as `dead` and `handle_sigint` not? Should be consistent in the source ...

mptre:
Lets keep it consistent, merged. Thanks!

PR #211, opened 2017-07-12
Add tty_size and Ctrl-L

DBOTW:


pick.c
```diff
@@ -11,6 +11,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/ioctl.h>
```
mptre: Should be placed on a separate line before the other includes, see [style(9)](http://man.openbsd.org/style).

pick.c
```diff
@@ -687,6 +695,24 @@ tty_restore(void)
 }
 
 void
+tty_size(void)
+{
+	struct winsize ws;
```
mptre: Indent using a single tab after `winsize`.

pick.c
```diff
@@ -687,6 +695,24 @@ tty_restore(void)
...
+
+	if (ioctl(fileno(tty_in), TIOCGWINSZ, &ws) != -1) {
+		tty_columns = ws.ws_col;
+		tty_lines = ws.ws_row;
+	} else {
+		if ((tty_columns = tigetnum("co")) < 1)
```
mptre: `co -> cols`

pick.c
```diff
@@ -687,6 +695,24 @@ tty_restore(void)
...
+			tty_columns = 80;
+		if ((tty_lines = tigetnum("li")) < 1)
```
mptre: `li -> lines`

mptre:
Great start, some inline comments. More general:

- The `COLUMNS` and `LINES` environment variables has to be respected. I
  tried to find an implementation that does resizing and found `vi(1)`
  which is defined by `POSIX`[1]. To summarize: the environment
  variables, if present, must have the highest precedence. This requires
  turing the values into integers, I would adovcate usage of
  `strtonum(3)`[2] from OpenBSD, this requires some
  autoconf/compat-tinkering that I could help out with.

  As for the implementation, you could look how `vi(1)`[3] does it.
  Instead of checking the environment variables first it's done as a
  last step in order to remove the need to if either `tty_columns` or
  `tty_lines` already is set when doing `TIOCGWINSZ` and `tigetnum()`. A
  simpler implementation IMHO with a negligible performance impact.

  This behavior reflect the current behavior of pick and is relied upon
  by by pick.vim[4].

- Add a test with something like the following:

```
description: CTRL-L clears the selection
keys: \016 \014 \\n # DOWN CTRL_L ENTER
stdin:
a
b
stdout:
a
```

- The commit message could focus more on the _what_ rather than
  reflecting the added code, something like:

  > Add redraw support with respect to TTY dimensions

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03
[2]: http://man.openbsd.org/strtonum
[3]: https://github.com/openbsd/src/blob/master/usr.bin/vi/cl/cl_term.c#L392-L420
[4]: https://github.com/calleerlandsson/pick.vim

DBOTW:
Closing in favour of https://github.com/calleerlandsson/pick/pull/212 ...

PR #212, opened 2017-07-16
Add resize of TTY by SIGWINCH signal

DBOTW:
Very simple approach to solve https://github.com/calleerlandsson/pick/issues/115, but works here (so far) without any flaws ...

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -41,6 +41,7 @@ enum key {
 	CTRL_A,
 	CTRL_E,
 	CTRL_K,
+	CTRL_L,
 	CTRL_U,
 	CTRL_W,
 	UP,
@@ -359,6 +360,9 @@ selected_choice(void)
 			query_length = cursor_position;
 			filter_choices();
 			selection = yscroll = 0;
+			break;
+		case CTRL_L:
+			tty_size();
 			break;
 		case CTRL_W:
 			if (cursor_position == 0)
@@ -883,6 +887,7 @@ get_key(char *buf, size_t size, size_t *nread)
 		KEY(CTRL_A,	"\001"),
 		KEY(CTRL_E,	"\005"),
 		KEY(CTRL_K,	"\013"),
+		KEY(CTRL_L,	"\014"),
 		KEY(CTRL_U,	"\025"),
 		KEY(CTRL_W,	"\027"),
 		CAP(DEL,	"kdch1"),
--- a/pick.c
+++ b/pick.c
@@ -2,6 +2,8 @@
 #include "config.h"
 #endif
 
+#include <sys/ioctl.h>
+
 #include <ctype.h>
 #include <err.h>
 #include <limits.h>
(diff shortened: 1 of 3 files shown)


pick.c
```diff
@@ -649,12 +608,72 @@ tty_init(void)
...
+#endif
+}
+
+void tty_draw(void)
+{
+	size_t	i, j;
+	size_t	xscroll = 0;
+
+	tty_putp(cursor_invisible, 0);
+	tty_putp(cursor_home, 0);	/* move cursor home to avoid screen clutter */
```
calleerlandsson: What kind of clutter appears if we don't output this? I tried removing this line but could reproduce any issues.

pick.c
```diff
@@ -649,12 +608,72 @@ tty_init(void)
...
+	tty_putp(carriage_return, 1);	/* move cursor to first column */
+	if (cursor_position >= xscroll + tty_columns)
+		xscroll = cursor_position - tty_columns + 1;
+	if (cursor_position < xscroll)
+		xscroll = cursor_position;
+	print_line(&query[xscroll], query_length - xscroll, 0, -1, -1);
+
+	if (selection - yscroll >= choices_lines)
+		yscroll = selection - choices_lines + 1;
```
calleerlandsson: This is off by one from the code that was removed from the `PAGE_DOWN` and `END` branches in `selected_choice()`. Is this because it is now done prior to printing the choices?

pick.c
mptre: This does not work with the alternate screen disabled (`-X`).

mptre:
Hi,
I appreciate the effort but I think we should go ahead and solve this is
step by step, your previous PR is a good first step. Please checkout
#213.

As for `SIGWINCH`, the only place where it should be allowed is during
the first call to `read(3)` in `get_key`. This approach will (hopefully)
require less churn. I'm willing to give a try once #213 is in.

pick.c
DBOTW: Sometimes after (rapid) switching between fullscreen and resized window, the choices appear in the bottom half of the fullscreen window, not on top. Pressing a key fixes this, so better remove this line to keep `alternate screen disabled` working ...

pick.c
DBOTW: Yes, but the reason was to prevent that the `selection` becomes "invisible" if the window was resized. It also fixes https://github.com/calleerlandsson/pick/issues/199. Instead of `yscroll = selection - choices_lines + 1;` it could (or should?) be `yscroll = selection - tty_lines` ...

pick.c
DBOTW: Sorry for the noise: this was a bug in `rxvt-unicode` version 9.20, update to 9.22 solves this ...

DBOTW:
OK, another approach (https://github.com/DBOTW/pick/commit/f08d779b6a481de32e8fd3e1f2811c781c1eb2c2) with less "churn" in the source, but breaks most tests ... is using `select` the "right way" to do it? Any comments/help on this will be very appreciated ...

mptre:
Nice, I have something similar in my tree. Checkout the [sigwinch]
branch.

[sigwinch]: https://github.com/calleerlandsson/pick/tree/sigwinch

DBOTW:
Great! This should definitely be merged!

mptre:
Can we close this in favor of #213 by now?

DBOTW:
Of course!

PR #213, opened 2017-07-17
SIGWINCH support

mptre:


mptre:
Any objections on using C99? /cc @calleerlandsson @mike-burns

DBOTW:
This will break my beloved `gcc 2.95` on ancient machines here ... but anyway: time moves on ... and all this just for including some stuff from OpenBSD-folks, that no one uses ... ;-)

mptre:
@DBOTW can you confirm that compiling this PR with GCC 2.95 still works?

DBOTW:
After some more "tweaking" (replace `-Wextra` with `-Wno-format`) and defining `LLONG_MIN` and `LLONG_MAX` in `compat-strtonum.c` (I knew this would cause trouble ;-) it works!

mptre:
@DBOTW thanks for confirming.

mptre:
The remaining pieces to get `SIGWINCH` working is now merged into this
PR.

Does anyone have access to a machine where `SIGWINCH` is not defined?
I'm more in favor of skipping the autoconf check if we can't test it
properly. Even the [tmux portable][tmux] assumes its present as far as I see it.

[tmux]: https://github.com/tmux/tmux/search?q=SIGWINCH

DBOTW:
Hard to find such a machine: even HP-UX 8.x (and earlier) have `SIGWINCH` (called `SIGWINDOW`) ...
I fired up my HP9000 Model 715/50 with HP-UX 10.20 and compiled `pick` with `SIGWINCH` support and it worked ... most of the time ... sometimes pressing `^L` is necessary to clean up the mess ... :-)

mptre:
Thanks for checking! I then advocate on omitting the autoconf check for
now and get this released. If we receive any reports on build failures
we could revisit this matter when we know it's an actual issue.

calleerlandsson:
This looks great @mptre and @DBOTW. Thanks for all your great work! Feel free to rebase and merge, @mptre.

mptre:
Thanks everyone, merged!

PR #215, opened 2017-07-29
More verbose error messages

DBOTW:


mptre:
This is actually intentional and a common idiom in OpenBSD, see commit
f5fde2d5739457953d4b04e056e7bc6d722ce8b8. The reasoning goes somewhat
along the lines as follows: if you managed to exhaust the memory you got
bigger problems and knowing which function that returned `ENOMEM` does
not add much value.

DBOTW:
Sorry for the noise: I should take a look into the `pick` git log and think before pulling a request ... ;-)

PR #216, opened 2017-07-30
Add support for NetBSD-curses

DBOTW:
NetBSD uses `curses`: `tparm` with 9 parameters and `UP` already defined in `termcap.h`
(Tested with NetBSD 7.1 ... NetBSD 5.1 lacks some definitions in `term.h` and breaks ...)

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -9,6 +9,11 @@
 #define __dead __dead2
 #endif /* __FreeBSD__ */
 
+#ifdef __NetBSD__
+#define tparm(a, b)	tparm(a, b, 0, 0, 0, 0, 0, 0, 0, 0)
+#define UP		_UP
+#endif
+
 #if defined(__linux__) || defined(__CYGWIN__)
 #ifndef __dead
 #ifdef __GNUC__


mptre:
Thanks for taking time to make pick work on other BSDs. Are you using
NetBSD as your daily driver? If that's the case consider making pick
available as a package on NetBSD. The work by the other package
maintainers have help us find bugs and other issues related to
portability and is of significant value to the project.

> NetBSD uses `curses`: `tparm` with 9 parameters

Please try out #218 and see if it solves the problem.

> ... and `UP` already definied in `termcap.h`

I would prefer if we instead renamed/prefixed `{UP,DOWN,RIGHT,LEFT}` to
not cause any conflicts rather than using yet another `ifdef`. Renaming
ideas are much appreciated.

DBOTW:
Sorry, NetBSD will never be my daily driver: too many things are broken there ... ;-)
What about `CURSOR_{UP,DOWN,RIGHT,LEFT}`?

mptre:
`UP` and `DOWN` affects the selection while `LEFT` and `RIGHT` the
cursor.

How about `LINE_UP` and `LINE_DOWN`similiar to `PAGE_{UP,DOWN}`.

mptre:
@DBOTW please try out #222.

DBOTW:
Works, thanks @mptre! But seems that their `setupterm` is broken as it always reports 80x24 (`stty -a` shows the correct size of the terminal window). Tested via `ssh` and in `xterm` & `rxvt`; played with different types of `TERM`, used `resize` & `reset`: always 80x24 in `pick`. The `winsz`-branch fixes this ...

DBOTW:
The above bug (`setupterm`) is fixed in "NetBSD-current" ... so closing this PR ...

PR #217, opened 2017-07-30
Fix compile warning on isprint()-call

DBOTW:
On some platforms (NetBSD 7.1 with gcc 4.8.5) the following warning appears:
```
pick.c: In function 'get_key':
pick.c:887:3: warning: array subscript has type 'char' [-Wchar-subscripts]
   if (isprint(buf[0]))
   ^
```
and due to `-Werror` the compilation exits ... this commit fixes it ...

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -884,7 +884,7 @@ get_key(char *buf, size_t size, size_t *nread)
 	}
 
 	if (!isu8start(buf[0])) {
-		if (isprint(buf[0]))
+		if (isprint((unsigned char)buf[0]))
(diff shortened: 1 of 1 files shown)


mptre:
Nice, this is correct. However, I have another diff sitting in my tree
that reworks the `get_key()` function and it solves this issue. If you
don't mind I would prefer getting that diff in very soon instead.

DBOTW:
OK, so `char buf[6];` will become `unsigned char buf[6];`? ;-)

mptre:
Correct.

PR #218, opened 2017-07-30
Always pass 9 parameters to tparm()

mptre:
Some implementations like NetBSD have a non-variadic prototype of
tparm() requiring 9 explicit parameters. Therefore always make sure to
pass 9 parameters. Since other implementations uses a variadic prototype
it will work on them as well.

Inspired by how this solved in the portable version of tmux.

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -80,6 +80,7 @@ static const char		*strcasechr(const char *, const char *);
 static int			 tty_getc(void);
 static const char		*tty_getcap(char *);
 static void			 tty_init(void);
+static const char		*tty_parm1(const char *, int);
 static int			 tty_putc(int);
 static void			 tty_restore(void);
 static __dead void		 usage(void);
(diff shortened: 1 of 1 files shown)


DBOTW:
It works, but `gcc (nb2 20150115) 4.8.5` on NetBSD 7.1 complains again:
```
pick.c:315:4: error: passing argument 1 of 'tty_parm1' discards 'const' qualifier from pointer target type [-Werror]
```
so it should be `tty_parm1(const char *, int)` ... and please rename it to `tty_parm` ... :-)

mptre:
> so it should be `tty_parm1(const char *, int)`

Well that doesn't work on OpenBSD since:

```c
char *tparm(char *, ...);
```

... and on NetBSD:

```c
char *tparm(const char *, ...);
```

It's not pretty but the updated commit should work, please confirm.

DBOTW:
Confirmed.

mptre:
Thanks, merged.

PR #221, opened 2017-07-31
Prevent selection bar vanishes

DBOTW:
better fix for issue https://github.com/calleerlandsson/pick/issues/199 that also works well if the terminal window will be resized (PR https://github.com/calleerlandsson/pick/pull/213)

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -304,6 +304,8 @@ selected_choice(void)
 		else
 			xscroll = 0;
 		print_line(&query[xscroll], query_length - xscroll, 0, -1, -1);
+		if (selection - yscroll >= choices_lines)
+			yscroll = selection - choices_lines + 1;
 		choices_count = print_choices(yscroll, selection);
 		if (choices_count - yscroll < choices.length
 		    && choices_count - yscroll < choices_lines) {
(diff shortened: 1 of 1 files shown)


mptre:
Thanks, I have to take a closer look at this later.

mptre:
@DBOTW Is this a bug fix or an attempt to reduce duplicated logic?

DBOTW:
@mptre it's both and resetting the selection on `CTRL-L` can be deleted:
```C
--- pick.c.orig	2017-08-24 19:08:12.452959404 +0200
+++ pick.c	2017-08-24 19:08:35.116958909 +0200
@@ -407,7 +407,6 @@
 			break;
 		case CTRL_L:
 			tty_size();
-			selection = yscroll = 0;
 			break;
 		case CTRL_W:
 			if (cursor_position == 0)
```
... please try it by resizing the window vertically lesser than the selection bar ...

mptre:
> please try it by resizing the window vertically lesser than the selection bar

Sorry for not getting back sooner. I'm having a hard time reproducing
any bugs related screen resizing and selection now that the selection is
being reset on resize. Is this still an issue or resolved?

DBOTW:
> now that the selection is being reset on resize

IMHO this is confusing and should be deleted, but if we do this we will get this behavoir:
old
... the selection bar vanishes ... with this PR it looks like:
new
... the selection bar is always visible on screen ...

So this also fixes https://github.com/calleerlandsson/pick/issues/199 by removing the range check on `Page-Down` and `End` ...

mptre:
Thanks for the GIFs, this sure looks like an improvement. What do use to
record your screen?

DBOTW:
> What do use to record your screen?

https://github.com/phw/peek

mptre:
Thanks! Tweaked and committed as 58c5b46.

PR #223, opened 2017-08-04
Add Escape key to exit

DBOTW:
Approach to fix issue https://github.com/calleerlandsson/pick/issues/125 with ideas from @mptre (https://github.com/calleerlandsson/pick/issues/125#issuecomment-222808260): works here and all tests passed.

### DIFF
--- a/pick.1
+++ b/pick.1
@@ -48,7 +48,7 @@ Disable the use of the alternate screen terminal feature.
 .El
 .Sh COMMANDS
 .Bl -tag -width XXXX
-.It Ic Ctrl-C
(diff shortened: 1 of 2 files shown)


mptre:
Thanks for taking the time to review and solve the open issues. However,
I'm not in favor of adding support for `ESC` since it's not too common
among other Unix utilities, there's also a more extensive discussion in
the issue.

DBOTW:
Right ... I don't use `ESC` but keep this in my tree if someone is interested ...

mptre:
Great! You could add a comment to the issue. Even though it's already
referenced a comment might be clearer.

PR #224, opened 2017-08-05
Use compound-list for POSIX compatibility

DBOTW:
... or some shells (`sh` on NetBSD) will fail to run the tests ...


mptre:
Thanks, merged as 716fc0a.

PR #225, opened 2017-08-05
Add ending semicolon as work-around for NetBSD sh

DBOTW:
... without this all checks fail with `Syntax error: ";;" unexpected` ... don't know why ...

### DIFF
--- a/tests/pick-test.sh
+++ b/tests/pick-test.sh
@@ -25,8 +25,8 @@ for testcase; do
       eval "${key}='${val%%#*}'"
     else
       case "$key" in
-      stdin)  tmpfile=$stdin; >$tmpfile ;;
-      stdout) tmpfile=$stdout; >$tmpfile ;;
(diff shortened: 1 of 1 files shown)


mptre:
The current code is indeed not compliant to what's stated in the sh(1)
manual (only checked the one on OpenBSD). I'd prefer your previous
solution using a command list enclosed in `{}`.

Some excerpts from sh(1):

> Command lists, as described above, can be enclosed within () to have
> them executed in a subshell, or within {} to have them executed in the
> current environment:
>
>   (command ...)
>   {command ...;}

...

> A case conditional is used to run commands whenever a pattern is matched.
> Its format is:
>
>   case word in
>    (pattern [| pattern ...]) command;;
>    ...
>   esac
>
> ... Multiple commands may be given by grouping them in lists, as
> described above, or by separating them with newlines.

DBOTW:
Thanks for merging the previous solution!

PR #226, opened 2017-08-06
Add function to replace query with selection

DBOTW:
`echo "item a\nitem b\nitem b 1\nitem b 2\nitem b 3"| pick`

with this function selecting `item b` and pressing `TAB` filters the choices accordingly ...


mptre:
Thanks for your proposal. I haven't felt this need myself, any thoughts from the others?

calleerlandsson:
Thanks for contributing, @DBOTW!

I can't say I've felt the need for this myself either. Could you tell us a bit about your use-case, @DBOTW?

While this diff is quite small, I think it's important that we're conservative when it comes to introducing new features or we might end up with a lot of added complexity.

DBOTW:
The idea came from `dmenu` (http://tools.suckless.org/dmenu), `fzy` (https://github.com/jhawthorn/fzy) calls it `autocomplete`: the selected choice can be easily modified before pressing `Alt-Enter`. But the main reason for this is: if I press the `Tab`-key out of habit something should happen ... ;-)

mptre:
While this being a neat idea, I'm favor of closing this for now until
enough compelling use-cases has been presented. OK to close?

DBOTW:
OK

PR #227, opened 2017-08-11
Pick elements without interface

luizperes:
I was using `pick` and realized that we don't always want to use it in interface mode, such as when you are including it in a script, thus:

This PR includes the flags:
  - `--first` (or `-e`): picks the first element of the list and exits
  -  `--last` (or `-E`): picks the last element of the list and exits
  - `--nth <number>` (or `-n <number>`): picks the element number `n` and exits _// the first element is the `nth` element `n=1`, the last element is the `nth` element `n=count`

Also, this PR prepares `pick` to `--long-options`, not sure if you want to use it, though.

The video below shows it better:

ezgif com-video-to-gif 1

_// Obs.: In spite of reading your coding spacement guide, I am not so sure about if the spaces are alright_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# Swap files
+*.swo
+*.swp
+
 *.o
 core*
 pick
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,7 +16,7 @@ TESTS=tests/01.t tests/02.t tests/03.t tests/04.t tests/05.t tests/06.t \
       tests/37.t tests/38.t tests/39.t tests/40.t tests/41.t tests/42.t \
       tests/43.t tests/44.t tests/45.t tests/46.t tests/47.t tests/48.t \
       tests/49.t tests/50.t tests/51.t tests/52.t tests/53.t tests/54.t \
-      tests/55.t tests/56.t
+      tests/55.t tests/56.t tests/57.t tests/58.t tests/59.t
 TEST_EXTENSIONS=.t
 T_LOG_COMPILER=$(top_srcdir)/tests/pick-test.sh
 check_PROGRAMS=tests/pick-test
--- a/pick.c
+++ b/pick.c
@@ -84,7 +84,7 @@ static int			 isu8cont(unsigned char);
 static int			 isu8start(unsigned char);
 static size_t			 min_match(const char *, size_t, ssize_t *,
 				    ssize_t *);
-static int			 print_choices(int, int);
+static int			 print_choices(int, int, int);
 static void			 print_line(const char *, size_t, int, ssize_t,
 				    ssize_t);
 static const struct choice	*selected_choice(void);
@@ -96,7 +96,7 @@ static const char		*tty_parm1(const char *, int);
 static int			 tty_putc(int);
 static void			 tty_restore(void);
 static __dead void		 usage(void);
-static void                      print_element_at_index(void);
+static const struct choice       *get_choice_at_index(void);
 
 static struct termios	 original_attributes;
 static struct {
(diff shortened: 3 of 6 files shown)


luizperes:
Will fix for ISO C90

mptre:
First of, thanks for your contribution and extensive description
including a GIF(!).

> I was using `pick` and realized that we don't always want to use it in interface mode, such as when you are including it in a script, thus:

While pick is not intended for use in non interactive environments, I
recognize the problem you're trying to solve. The behavior you're
seeking could be solved using the traditional suite of Unix utilities.
Although, your not able to use the fuzzy search algorithm provided by
pick but since you're scripting, I assume what you're looking for could
be expressed using a regular expression. Below are some proposals on
how-to replicate the behavior:

> This PR includes the flags:
>   - `--first` (or `-e`): picks the first element of the list and exits

```sh
command | grep query | head -1
command | grep query | sed -n '1p'
```

>   -  `--last` (or `-E`): picks the last element of the list and exits

```sh
command | grep query | tail -1
command | grep query | sed -n '$p'
```

>   - `--nth <number>` (or `-n <number>`): picks the element number `n` and exits _// the first element is the `nth` element `n=1`, the last element is the `nth` element `n=count`

```sh
command | grep query | tail -n +<number> | head -1
command | grep query | sed -n '<number>p'
```

> Also, this PR prepares `pick` to `--long-options`, not sure if you want to use it, though.

GNU is not encouraged here unfortunately. Although I'm not to keen on
merging this, here's some general feedback on the code:

- Removing included headers might work locally but break on other
  Operating Systems. Therefore make sure to preserve all included
  headers.

- Prefer using a global gitignore for your own opinionated ignores. See
  `git-config(5)` and search for `core.excludesFile`.

- Regarding indentation etc., see the newly updated documentation on
  [contributing]. Some nits below:

  - Lines should not be longer than 80 characters

  - Some instances of using spaces prior tabs for indentation

I hope your not discouraged by my response and keep contribution to
pick. Thanks for taking the time and effort to submit this PR.

[contributing]: https://github.com/calleerlandsson/pick/blob/master/DEVELOPING.md

DBOTW:
If the fuzzy search algorithm of `pick` is needed: https://github.com/DBOTW/pick/commit/f1d83c5fc3f8898f4984f0c8040a803605b546a7:
```sh
command | pick -e -q query | head -1
```
to pick the first element of the list for example ...

luizperes:
Hi @mptre! Sorry for delayed response.
Now I understand better your guidelines and the purpose of `pick`, however I still think that it would be great if such possibility existed. When I was looking for a solution to my problem, the word `pick` suggested (in my opinion) that I could pick an element over a list, thus I installed and realized that it wasn't as handy as it could be, or at least as I thought it could be, despite of the use of `grep` or not for such tasks (which I didn't know/pay attention to my search). After I found `pick`, it just felt I was in the right track and this is the reason for my PR.

I believe that what @DBOTW wrote is very interisting and it feels that it matches with the way `pick` should/could work in non-interface mode. I vote for that. What do you think?

For everything else, I really understand all of your concerns and will improve in a future contribution.

Best,

mptre:
While @DBOTW solution is neat I would still argue that there are better
tools available for picking lines in a non-interactive context.

luizperes:
I actually agree with you (now), @DBOTW, should I close this PR?

calleerlandsson:
I really appreciate your proposal, @luizperes, and I'm grateful for you putting time and effort into contributing to pick.

For now, I'd like pick to remain focused on supporting interactive use.

Thanks and welcome to the project!

luizperes:
Thanks @calleerlandsson!

PR #228, opened 2017-08-12
Tag or untag selected choices

DBOTW:
... simple tagging-mechanism:
tagged

### DIFF
--- a/pick.1
+++ b/pick.1
@@ -60,9 +60,11 @@ selected choice.
 .It Ic Home/End | Alt-</Alt->
 Move the selection to the first/last choice matching the current search query.
 .It Ic Enter
-Output the currently selected choice and exit.
+Output the currently selected choice or tagged choice/s and exit.
 .It Ic Alt-Enter
 Output the current input query and exit.
+.It Ic Ins | Ctrl-Space
+Tag or untag the currently selected choice.
 .It Ic Left/Right | Ctrl-B/Ctrl-F
 Move the cursor left and right in the search query input field.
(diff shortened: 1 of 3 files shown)


calleerlandsson:
Hey @DBOTW, cool proposal! Thanks for the great GIF showing how it works!

I'm a bit hesitant towards merging this for the same reasons as #226. Could you tell us a bit more about your use-case?

What are your thoughts, @mptre?

DBOTW:
Thanks @calleerlandsson, saw this is in other `fuzzy finders` and thought it would be "amusing" to move around and select stuff from the list ... I'm not really expecting this to be merged ... think of it as a "playground", although it can be useful for selecting filenames and pass them to an application/command or something like that ... I fully understand that `pick` should be simple and not overloaded with such "features", but I also wanted to post a GIF that looks cool ... ;-)

mptre:
> What are your thoughts, @mptre?

Neat idea indeed but I haven't felt the need myself for such a feature.

calleerlandsson:
I think this feature is cool and can definitely see how it could be useful but until we have a strong use-case, I'd rather hold off on merging this in favor of lower complexity.

I want to emphasize that I don't wish to discourage you from contributing to pick. I'm also happy to discuss this feature again in the future.

DBOTW:
No worries, @calleerlandsson: I'm always happy to contribute to `pick` as a hobbyist with no professional background ... :-) I'm also very grateful, that @mptre looks critical over my PRs and points me to alternative (mostly obviously) ways to solve the problem ...

PR #229, opened 2017-08-12
Treat slash as word-delimiter in Ctrl-W

DBOTW:
... otherwise a very long pathname in the query will be  completely deleted ...


mptre:
I'm in favor such a change. Ideally, we would implement the `ALTWERASE`
algorithm treating `!isalpha(3)` as delimiters.

PR #232, opened 2017-08-25
Set number of choices lines to display

DBOTW:
... sometimes a full screen list of choices is not desired (with disabled alternate screen):
numlines
while testing the code the cursor sometimes moves to home position: as the following fixes this
```diff
@@ -318,7 +324,7 @@ selected_choice(void)
                        if (tty_putc('\n') == EOF)
                                err(1, "tty_putc");
                        tty_putp(clr_eos, 1);
-                       tty_putp(tty_parm1(parm_up_cursor, choices_count + 1),
+                       tty_putp(tty_parm1(parm_up_cursor, choices_count - yscroll + 1),
                            1);
                } else if (choices_count > 0) {
                        /*
```
I think it's a "bug" in the original code ...

### DIFF
--- a/pick.1
+++ b/pick.1
@@ -9,6 +9,7 @@
 .Op Fl hvS
 .Op Fl d Op Fl o
 .Op Fl x | Fl X
+.Op Fl l Ar lines
 .Op Fl q Ar query
 .Sh DESCRIPTION
 The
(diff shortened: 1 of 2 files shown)


mptre:
You could use the `LINES` environment variable instead:

```sh
$ git ls-files | env LINES=10 pick -X | xargs less
```

Thanks for the bug report, fixed in 50e2748.

PR #233, opened 2017-08-27
Favor the shortest match

mptre:
Just discovered a severe bug in the matching algorithm, please give it a
try:

> Instead of unconditionally returning the length of the current match,
> make sure the shortest match bubbles up along the recursive call stack.
> 
> Note that this bug only occurs when a choice contains multiple instances
> of the query and the right-most match is the shortest one. See the
> example below including the choice, query, match prior this commit and
> after:
> 
> ```
> Choice       Query  Before       After
> aa/åå/aa/aa  aa/aa  aa/åå/aa/aa  aa/åå/aa/aa
> 		    ^      ^           ^   ^
> 		    |------|           |---|
> ```


calleerlandsson:
Good catch, @mptre! With your test in place we should also be able to better avoid regressions.

Looks good to merge.

mptre:
Thanks, merged.

PR #235, opened 2017-08-30
Fix: use tty_columns for width in xscroll

DBOTW:


mptre:
Nice catch, merged as 6ce2b56.

PR #236, opened 2017-08-31
Refactor get_key()

mptre:
Simplify the prototype by only passing a pointer-pointer that will point
to the input buffer upon return.

While here, declare the buffer as unsigned since the signedness of char
can differ between platforms. This issue was also pointed out by Jenz
Guenther in PR #217.

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -70,7 +70,7 @@ static void			 delete_between(char *, size_t, size_t, size_t);
 static char			*eager_strpbrk(const char *, const char *);
 static void			 filter_choices(void);
 static char			*get_choices(void);
-static enum key			 get_key(char *, size_t, size_t *);
+static enum key			 get_key(const char **);
 static __dead void		 handle_sigint(int);
 static void			 handle_sigwinch(int);
 static int			 isu8cont(unsigned char);
@@ -286,11 +286,11 @@ eager_strpbrk(const char *string, const char *separators)
 const struct choice *
 selected_choice(void)
 {
-	size_t	cursor_position, i, j, length, xscroll;
-	char	buf[6];
-	int	choices_count;
-	int	selection = 0;
-	int	yscroll = 0;
+	const char	*buf;
+	size_t		 cursor_position, i, j, length, xscroll;
+	int		 choices_count;
+	int		 selection = 0;
+	int		 yscroll = 0;
 
 	cursor_position = query_length;
 
@@ -345,7 +345,7 @@ selected_choice(void)
 		tty_putp(cursor_normal, 0);
 		fflush(tty_out);
 
-		switch (get_key(buf, sizeof(buf), &length)) {
+		switch (get_key(&buf)) {
 		case ENTER:
 			if (choices_count > 0)
 				return &choices.v[selection];
@@ -490,6 +490,8 @@ selected_choice(void)
 			yscroll = selection = 0;
 			break;
 		case PRINTABLE:
+			length = strlen(buf);
+
 			if (query_length + length >= query_size) {
 				query_size = 2*query_length + length;
 				if ((query = reallocarray(query, query_size,
@@ -888,7 +890,7 @@ print_choices(int offset, int selection)
 }
 
 enum key
-get_key(char *buf, size_t size, size_t *nread)
+get_key(const char **key)
 {
 #define	CAP(k, s)	{ k,	s,	NULL,	0 }
 #define	KEY(k, s)	{ k,	NULL,	s,	sizeof(s) - 1 }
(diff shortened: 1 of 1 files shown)


pick.c
```diff
@@ -928,16 +930,20 @@ get_key(char *buf, size_t size, size_t *nread)
 		KEY(RIGHT,	"\006"),
 		KEY(UNKNOWN,	NULL),
 	};
-	int	c, i;
+	static unsigned char	buf[8];
```
calleerlandsson: Out of curiosity: What are the pros of scoping this static variable to `get_key` and letting `*key` point to it rather than declaring it outside `get_key` and letting `selected_choice` refer directly to it? Making it possible to declare it `const`?

The size of `buf` seems to have increased from `6` to `8`. What was the reason for doing that?

mptre: > Out of curiosity: What are the pros of scoping this static variable to `get_key` and letting `*key` point to it rather than declaring it outside `get_key` and letting `selected_choice` refer directly to it? Making it possible to declare it `const`?

Less dereferencing going on in `get_key` and if we where to declare
`buf` as `unsigned` in `selected_choice` it would require lots of
casting while passing it to various functions. I find the `get_key` code
a bit easier to follow after applying this diff, but that's just my
opinion.

> The size of `buf` seems to have increased from `6` to `8`. What was the reason for doing that?

No real reason, other than 8 being a power of 2.

pick.c
```diff
@@ -928,16 +930,20 @@ get_key(char *buf, size_t size, size_t *nread)
...
+	len = 0;
 
 	/*
 	 * Allow SIGWINCH on the first read. If the signal is received, return
 	 * CTRL_L which will trigger a resize.
 	 */
 	toggle_sigwinch(1);
-	buf[(*nread)++] = tty_getc();
+	buf[len++] = tty_getc();
+	buf[len] = '\0';
```
DBOTW: How about using `memset` at the beginning to zero out the buffer?

mptre: Interesting, I wonder if Coverity will complain though: 3e04320.

DBOTW:
```diff
diff --git a/pick.c b/pick.c
index 73cc0e4..194db1d 100644
--- a/pick.c
+++ b/pick.c
@@ -930,7 +930,7 @@ get_key(const char **key)
                KEY(RIGHT,      "\006"),
                KEY(UNKNOWN,    NULL),
        };
-       static unsigned char    buf[8];
+       static unsigned char    buf[8] = {'\0'};
        size_t                  len;
        int                     c, i;
 
@@ -943,7 +943,6 @@ get_key(const char **key)
         */
        toggle_sigwinch(1);
        buf[len++] = tty_getc();
-       buf[len] = '\0';
        toggle_sigwinch(0);
        if (gotsigwinch) {
                gotsigwinch = 0;
@@ -971,7 +970,6 @@ get_key(const char **key)
                if (len == sizeof(buf) - 1)
                        break;
                buf[len++] = tty_getc();
-               buf[len] = '\0';
        }
 
        if (len > 1 && buf[0] == '\033' && (buf[1] == '[' || buf[1] == 'O')) {
@@ -1006,7 +1004,6 @@ get_key(const char **key)
                        return UNKNOWN;
 
                buf[len++] = tty_getc();
-               buf[len] = '\0';
        }
 
        return PRINTABLE;
```

DBOTW:
```diff
diff --git a/pick.c b/pick.c
index 73cc0e4..5354ea2 100644
--- a/pick.c
+++ b/pick.c
@@ -937,13 +937,14 @@ get_key(const char **key)
        *key = (const char *)buf;
        len = 0;
 
+       memset(buf, '\0', sizeof(buf));
+
        /*
         * Allow SIGWINCH on the first read. If the signal is received, return
         * CTRL_L which will trigger a resize.
         */
        toggle_sigwinch(1);
        buf[len++] = tty_getc();
-       buf[len] = '\0';
        toggle_sigwinch(0);
        if (gotsigwinch) {
                gotsigwinch = 0;
@@ -971,7 +972,6 @@ get_key(const char **key)
                if (len == sizeof(buf) - 1)
                        break;
                buf[len++] = tty_getc();
-               buf[len] = '\0';
        }
 
        if (len > 1 && buf[0] == '\033' && (buf[1] == '[' || buf[1] == 'O')) {
@@ -1006,7 +1006,6 @@ get_key(const char **key)
                        return UNKNOWN;
 
                buf[len++] = tty_getc();
-               buf[len] = '\0';
        }
 
        return PRINTABLE;
```

mptre: Coverity didn't complain so this change is in, thanks!

DBOTW:
Tested on `Debian testing` and `NetBSD 7.1`: so it should be merged ... ;-)

mptre:
@DBOTW thanks, merged.

PR #237, opened 2017-09-01
Unify mbtowc() invocation and error handling

mptre:
An abstraction worth performing since we can unify the internal state
reset and there's currently no need to distinguish between the two
scenarios where mbtowc() returns -1 or 0.


mptre:
If noone objects I'm going to go ahead and merged this soon.

calleerlandsson:
I think this looks great!

mptre:
Thanks @calleerlandsson, merged.

PR #239, opened 2017-09-11
Consistently use size_t while counting choices

mptre:
- Since we can have SIZE_T_MAX many choices, don't use plain signed
  integers.

- Make `choices_lines` unsigned to keep it consistent with
  `tty_{columns,lines}` since commit b5fe77f ("Make tty_{columns,lines}
  unsigned").

- Drop (by now) redundant casts to size_t

- Rework some logic in order to prevent doing subtraction on unsigned
  integers since it could result in wrapping.

- Rework the complex loop in print_choices() by keeping the upper bound
  in the loop declaration. The other loop condition is better suited
  inside the body of the loop. This convention is in general a good rule
  of thumb while constructing complex loops.

Testing would be much appreciated! Especially `Page-{Down,Up}`.

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -78,7 +78,7 @@ static int			 isu8start(unsigned char);
 static int			 isword(const char *);
 static size_t			 min_match(const char *, size_t, ssize_t *,
 				    ssize_t *);
-static int			 print_choices(int, int);
+static size_t			 print_choices(size_t, size_t);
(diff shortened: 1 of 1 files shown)


DBOTW:
Looks good on `Debian testing` & `NetBSD 7.1` ...

mptre:
Thanks @DBOTW, merged.

PR #240, opened 2017-09-11
Explicitly handle control characters

mptre:
The current SIGINT handler is not considered safe since it invokes
functions that are not considered asynchronous safe[1]. Quoting the
CERT C Coding Standard:

  In general, it is not safe to invoke I/O functions from within signal
  handlers.

Instead, do not turn control characters into signal but instead handle
the relevant ones explicitly. A pleasant side-effect is being able to
test the Ctrl-C error path.

While here, fix the broken suspend/resume behavior by resetting the
initial terminal state prior suspending and re-read it upon resume. I do
actually use suspend/resume while running pick when I forgot to check
something that will affect the choice to select.

[1] https://www.securecoding.cert.org/confluence/display/c/SIG30-C.+Call+only+asynchronous-safe+functions+within+signal+handlers


DBOTW:
Great work! I hope we will see `pick-1.9.0` with this soon ...

teoljungberg:
I opened https://github.com/calleerlandsson/pick.vim/pull/21 to resolve this in pick.vim too.

PR #242, opened 2017-09-15
Fix broken INSTALL.md.in path while running make distcheck

mptre:
The distcheck target ensures that the distribution tarball is
self-contained, i.e. using only the files present in the distribution it
can create another identical [distribution]. While doing so, we can't
assume that the current directory is the top-level.

This makes `make distcheck` work for me on OpenBSD. /cc @mike-burns 

[distribution]: https://www.gnu.org/software/automake/manual/automake.html#Checking-the-Distribution


mptre:
Thanks, merged as 9034255.

PR #245, opened 2017-09-17
Simplify error handling of tty_putc

DBOTW:
`putc` returns `EOF` on errors, not `ERR` (although both are defined as `-1`) ...

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -318,8 +318,7 @@ selected_choice(void)
 			 * column will be also be cleared. Therefore, move down
 			 * one line before clearing the screen.
 			 */
-			if (tty_putc('\n') == EOF)
-				err(1, "tty_putc");
+			tty_putc('\n');
 			tty_putp(clr_eos, 1);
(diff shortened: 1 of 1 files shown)


mptre:
The `ERR -> EOF` looks good. `tty_putc` does not perform any error
handling since it's used with `tputs(3)`, quoting the manual:

> ... putc is a putchar-like routine to which the characters are passed,
> one at a time.

I don't know if it's advised to perform error handling inside `putc` if
`tputs` relies on getting `EOF` back and perform any internal cleanup.

DBOTW:
>  if `tputs` relies on getting `EOF`  back and perform any internal cleanup.

http://pubs.opengroup.org/onlinepubs/7908799/xcurses/putp.html says:
> The tputs() function ignores the return value of putfunc.

quoting `man putc` here:
> fputc(), putc() and putchar() return the character written as an unsigned char cast to an int or EOF on error.

mptre:
Well then, thanks! Tweaked and committed as 1585928.

PR #248, opened 2017-09-20
Ignore OSC escape sequences

mptre:
It was brought to my attention that pick didn't handle OSC escape
sequences correctly. Make sure to ignore them while calculating the line
width for each choice and ignore them while matching.


DBOTW:
Tested and should be merged ...

calleerlandsson:
This looks good to me!

PR #249, opened 2017-09-22
Fix build with musl libc

ebfe:
CCEQ is not available under musl libc.

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -28,6 +28,10 @@
 
 #include "compat.h"
 
+#ifndef CCEQ
+	#define CCEQ(val, c) ((c) == (val) && (val) != _POSIX_VDISABLE)
+#endif
+
 #define tty_putp(capability, fatal) do {				\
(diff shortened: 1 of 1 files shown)


mptre:
Thanks for the PR. I'd rather just inline the macro then, please try out
#250 and let me know if it works.

I'm trying to build with musl on [Travis] but ran into some issues. How
are you compiling pick with musl? Is anything more than the commands
below necessary? Travis uses Ubuntu and I did install the `musl-tools`
package.

```sh
$ env CC=musl-gcc ./configure
$ make
```

[Travis]: https://travis-ci.org/calleerlandsson/pick/jobs/278886966

ebfe:
#250 is fine too. I'm building on a Void Linux system with musl as the system libc. No other changes necessary here.

mptre:
Great, merged.

Are you packaging pick for Void Linux? If true, consider submitting a PR
with install instructions added to the README.

PR #251, opened 2017-09-23
Low-hanging optimization

mptre:
If the user didn't add more characters to the query all choices have to
be reconsidered as potential matches. In the opposite scenario, there's
no point in reconsidered all choices again since the ones that didn't
match the previous query will clearly not match the current one due to
the fact that previous query is a left-most substring of the current
one.

Testing would be much appreciated. On large inputs, pick should feel
snappier by now.

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -298,7 +298,8 @@ const struct choice *
 selected_choice(void)
 {
 	const char	*buf;
-	size_t		 choices_count, cursor_position, i, j, length, xscroll;
+	size_t		 cursor_position, i, j, length, xscroll;
+	size_t		 choices_count = 0;
 	size_t		 selection = 0;
 	size_t		 yscroll = 0;
 	int		 dofilter = 1;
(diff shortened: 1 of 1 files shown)


DBOTW:
`gcc (nb2 20150115) 4.8.5` on `NetBSD 7.1` shows:
```sh
pick.c: In function 'main':
pick.c:554:2: error: 'choices_count' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  for (i = 0; i < nchoices; i++) {
  ^
pick.c:300:11: note: 'choices_count' was declared here
  size_t   choices_count, cursor_position, i, j, length, xscroll;
```
... no such errors on `Debian` ...

mptre:
Thanks for testing @DBOTW. Could your try the latest commit on this
branch?

DBOTW:
@mptre: tested & works (GCC 2.95 also complained about this) ...

mptre:
Thanks, merged.

PR #254, opened 2017-09-24
Optimize redrawing of choices

DBOTW:
... each time the cursor moves only in the input query field all choices are printed again even though nothing has changed ... this PR avoids the unnecessary redraw and saves some CPU cycles ... ;-)

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -305,6 +305,7 @@ selected_choice(void)
 	size_t		 selection = 0;
 	size_t		 yscroll = 0;
 	int		 dofilter = 1;
+	int		 draw_choices = 1;
 	int		 query_grew = 0;
 
 	cursor_position = query_length;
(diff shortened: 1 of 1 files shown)


mptre:
Thanks. I agree in principle that this is more efficient but wonder if
it improves the responsiveness of the UI? Do you experience
flickering/rendering artefacts without this diff applied?

DBOTW:
Using `pick` over a serial terminal line (VT520 with 38400 baud rate) it improves cursor movement drastically ... believe me: you don't want a full screen redraw each time on such terminals ... ;-)

mptre:
Sorry for taking so long. See #264 for a slightly different approach,
reducing the necessary indent level a bit.

DBOTW:
Thanks, I'm closing this in favor of #264 ...

mptre:
Great, I'm planning on doing a release tomorrow.

PR #257, opened 2017-10-05
Add Void Linux install instructions

ebfe:


mptre:
Thanks, merged in commit dce2ddf.

PR #258, opened 2017-10-08
Add installation instruction for CRUX

rtlanceroad:


mptre:
Thanks, merged in commit 26eea32.

rtlanceroad:
Thanks! The Distro name is just `CRUX`, not `Crux`, each letter should be just capitalized . Could you fix it?

mptre:
Are you referring to the [reference]?

[reference]: https://github.com/calleerlandsson/pick/commit/26eea32febc2554543cf7b162d22f14ecd076a65#diff-04c6e90faac2675aa89e2176d2eec7d8R76

rtlanceroad:
Yeah, since it's just a reference, and I know that doesn't make much sense, but that is what I think the way it is.

mptre:
I went ahead and lowercase:d all references which makes more sense IMHO
since they are never literally displayed. See commit 3b69d1d.

PR #260, opened 2017-10-18
Improve tests

mptre:
The test suite has served us well by now. However, I've always found the
naming convention of the test files too obfuscated and it's hard to get
an overview of what's covered by the tests. Here's a first step allowing
multiple tests to be specified in a single file.


calleerlandsson:
This makes finding a specific test so much easier and will also help reduce conflicts between open PRs. Thanks, @mptre!

mptre:
Merged! Next up is doing a second pass going through the tests for
improved cleanup/consistency and making sure all supported keys/options
are tested. If anyone wants to help out, this is a perfect time.

PR #261, opened 2017-10-20
Emulate vi command mode

DBOTW:
... for https://github.com/calleerlandsson/pick/issues/146 ...

### DIFF
--- a/pick.1
+++ b/pick.1
@@ -9,6 +9,7 @@
 .Op Fl hvKS
 .Op Fl d Op Fl o
 .Op Fl x | Fl X
+.Op Fl y
 .Op Fl q Ar query
 .Sh DESCRIPTION
 The
@@ -52,6 +53,10 @@ Enable the use of the alternate screen terminal feature.
 This is the default.
 .It Fl X
 Disable the use of the alternate screen terminal feature.
+.It Fl y
+Start in
+.Xr vi 1
+command mode
 .El
 .Sh COMMANDS
 .Bl -tag -width XXXX
@@ -96,6 +101,12 @@ Delete to the beginning of the line in the search query input field.
 Delete to the end of the line in the search query input field.
 .It Ic Printable characters
 Added to the search query and will refine the current search.
+.It Ic Escape
+Toggle between the default
+.Xr emacs 1
+and
+.Xr vi 1
+command mode.
 .El
 .Sh ENVIRONMENT
 The following environment variables will affect the execution of
--- a/pick.c
+++ b/pick.c
@@ -6,6 +6,7 @@
 
 #include <ctype.h>
 #include <err.h>
+#include <errno.h>
 #include <limits.h>
 #include <locale.h>
 #include <poll.h>
@@ -73,6 +74,7 @@ static char			*eager_strpbrk(const char *, const char *);
 static void			 filter_choices(size_t);
 static char			*get_choices(void);
 static enum key			 get_key(const char **);
+static enum key			 get_vi_key(unsigned char);
 static void			 handle_sigwinch(int);
 static int			 isu8cont(unsigned char);
 static int			 isu8start(unsigned char);
@@ -111,6 +113,7 @@ static int			 descriptions;
 static int			 sort = 1;
 static int			 use_alternate_screen = 1;
 static int			 use_keypad = 1;
+static int			 vi_mode;
 
 int
 main(int argc, char *argv[])
@@ -128,7 +131,7 @@ main(int argc, char *argv[])
 		err(1, "pledge");
 #endif
 
-	while ((c = getopt(argc, argv, "dhoq:KSvxX")) != -1)
+	while ((c = getopt(argc, argv, "dhoq:KSvxXy")) != -1)
 		switch (c) {
 		case 'd':
 			descriptions = 1;
@@ -160,6 +163,9 @@ main(int argc, char *argv[])
 			break;
 		case 'X':
 			use_alternate_screen = 0;
+			break;
+		case 'y':
+			vi_mode = 1;
 			break;
 		default:
 			usage();
@@ -203,7 +209,7 @@ main(int argc, char *argv[])
 __dead void
 usage(void)
 {
-	fprintf(stderr, "usage: pick [-hvKS] [-d [-o]] [-x | -X] [-q query]\n"
+	fprintf(stderr, "usage: pick [-hvKS] [-d [-o]] [-x | -X] [-y] [-q query]\n"
 	    "    -h          output this help message and exit\n"
 	    "    -v          output the version and exit\n"
 	    "    -K          disable toggling of keyboard transmit mode\n"
@@ -212,6 +218,7 @@ usage(void)
 	    "    -o          output description of selected on exit\n"
 	    "    -x          enable alternate screen\n"
 	    "    -X          disable alternate screen\n"
+	    "    -y          start in vi command mode\n"
 	    "    -q query    supply an initial search query\n");
 
 	exit(1);
@@ -298,6 +305,7 @@ const struct choice *
 selected_choice(void)
 {
 	const char	*buf;
+	enum key	 key;
 	size_t		 cursor_position, i, j, length, xscroll;
 	size_t		 choices_count = 0;
 	size_t		 selection = 0;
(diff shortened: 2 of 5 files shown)


mptre:
Neat, less code than I expected. However, if we where to go down the
route of adding Vi-emulation my fear is that people would want more and
more key bindings supported from Vi; potentially ending up lots of code.
I think we already by now support a fine set of key bindings. What are
your thoughts @calleerlandsson?

Since you took the time to submit this, I'm going to review the code
anyway. In general, tests are missing. Should be even easier by now with
the updated test suite.

pick.1
```diff
@@ -8,6 +8,7 @@
 .Nm
 .Op Fl hvKS
 .Op Fl d Op Fl o
+.Op Fl V
```
mptre: Uppercase letters are by convention used to disable a feature in pick. I guess you chosed `V` since `v` already is taken. Enabling Vi-mode might therefore be done using a separate key binding.

pick.c
```diff
@@ -111,6 +112,7 @@ static int			 descriptions;
 static int			 sort = 1;
 static int			 use_alternate_screen = 1;
 static int			 use_keypad = 1;
+static int			 vi_mode = 0;
```
mptre: No need to initialize a static variable to 0. Might be done for clarity, though.

pick.c
```diff
@@ -973,6 +983,17 @@ get_key(const char **key)
 		return CTRL_L;
 	}
 
+	if (buf[0] == '\033') {
+		struct pollfd	pfd;
+
+		pfd.fd = fileno(tty_in);
+		pfd.events = POLLIN;
+		if (poll(&pfd, 1, 200) == 0) {
+			vi_mode = !vi_mode;
```
mptre: I would expect ESC to exit insert mode and not toggle Vi-mode.

```c
if (vi_mode)
  vi_mode = 0;
```

pick.c
```diff
@@ -1011,7 +1032,8 @@ get_key(const char **key)
 		 */
 		c = buf[len - 1];
 		while (c < '@' || c > '~')
-			c = tty_getc();
+			if (read(fileno(tty_in), &c, 1) == ERR)
```
mptre: This will break on Big Endian machines. Since `c` is declared as an `int` the read byte will end up in the Most-Significant-Byte (`c & 0xff000000`). The remaning bytes of `c` will be filled with potential garbage from stack since `c` is not initialized. Reading uninitialized memory from the stack is in addition undefined behavior. `c` should therefore be declared as `char` and preferably `unsigned char`. Note that this is only important when passing the argument to `read(3)`. The variable can then safely be promoted to a larger type such as `int`.

Prior to release, I always run the tests on an old Mac with a 32-bit PPC processor.

pick.c
```diff
@@ -1041,13 +1063,55 @@ get_key(const char **key)
 	return PRINTABLE;
 }
 
+enum key
+get_vi_key(const char **key)
```
mptre: Don't see the point of a pointer-pointer (no pun intended!). Probably due to copy-pasting the prototype for `get_key`.

pick.c
```diff
@@ -1041,13 +1063,55 @@ get_key(const char **key)
...
+{
+	static struct {
+		const char	*str;
+		enum key	 key;
+		int		 vi_mode;
+	}			keys[] = {
+		{ " ",		RIGHT,		1 },
```
mptre: I would rather see the bindings being defined inside `get_key` and adding a field to the struct indicating that some bindings are only applicable to Vi-mode.

pick.c
```diff
@@ -1041,13 +1063,55 @@ get_key(const char **key)
...
+
 int
 tty_getc(void)
 {
-	int	c;
+	int	c, n;
 
-	if ((c = getc(tty_in)) == ERR && !gotsigwinch)
-		err(1, "getc");
+	while ((n = read(fileno(tty_in), &c, 1)) == 0);
```
mptre: Same reasoning as above regarding the type of `c`.

pick.c
```diff
@@ -1041,13 +1063,55 @@ get_key(const char **key)
...
+	if (n == ERR && !gotsigwinch)
```
mptre: I would prefer `-1` instead of `ERR` for consistency with the rest of the code. Also, continuing reading on `errno == EINTR` would be a good addition.

pick.1
DBOTW: Didn't know about this convention and you guessed right ... ;-) I'm thinking of `-y` as the command line switch to enable the `vi`-mode at start ... a separate key binding would maybe inconvenient when `pick` is called from a script (`pick.vim`) ...

pick.c
DBOTW: Right ... fixed ...

pick.c
DBOTW: ... a separate key binding would then be necessary to enable the Vi-mode again ...

pick.c
DBOTW: Thanks for the clarification: I've fixed this. Would it make sense (although not necessary) to change it also in the original code?

pick.c
DBOTW: Of course that was the reason ... ;-)

pick.c
DBOTW: Changed it to:
```C
again:
        while ((n = read(fileno(tty_in), &c, 1)) == 0);
        if (n == -1) {
                if (errno == EINTR)
                        goto again;
                if (!gotsigwinch)
                        err(1, "read");
        }

        return c;
```
OK?

DBOTW:
@mptre: thank you very much for reviewing the code and the helpfull comments! I'm also satisfied with the current key bindings in `pick` but took the "issue" as an "exercise" ...

pick.c
mptre: Not necessary due to `int getc(FILE *)`.

pick.c
mptre: I would rather loop until success/error, untested code below:

I would rather loop until success/error, untested code below:

```c
for (;;) {
  n = read(fileno(tty_in), &c, 1);
  if (n == -1) {
    if (errno == EINTR)
      continue;
    if (gotsigwinch)
      break;
    err(1, "read");
  }
  if (n > 0)
    break;
}
```

mptre:
> I'm also satisfied with the current key bindings in `pick` but took the "issue" as an "exercise" .

That's the spirit!

pick.c
DBOTW: The idea was some sort of "hybrid-mode":  the "insert mode" is the "emacs mode" with all the known key bindings. By pressing `ESC` you switch to the Vi command-mode and the keys are "translated" to appropriate function of the "emacs mode": some commands (or `ESC`) brings you back to the "emacs mode". The command line switch only activates the Vi command-mode at start to mimic the behavior of Vi: you have to press some command keys (like `a`, `i`, `/`) or `ESC` first to enter the search query ... don't know if Vi-addicts like or want this ... ;-)

pick.c
```diff
@@ -371,10 +379,16 @@ selected_choice(void)
 			 * move the cursor if the position is non zero.
 			 */
 			tty_putp(tty_parm1(parm_right_cursor, j), 1);
-		tty_putp(cursor_normal, 0);
+		if (vi_mode)
+			tty_putp(cursor_visible, 0);
```
DBOTW: Although this will not work on all terminals, it seems the only "portable" way to change the cursor appearance (as an indicator that the Vi command-mode is active) ... the alternative is: https://vt100.net/docs/vt510-rm/DECSCUSR.html

pick.c
```diff
@@ -761,6 +775,7 @@ tty_restore(int doclose)
 
 	tty_putp(carriage_return, 1);	/* move cursor to first column */
 	tty_putp(clr_eos, 1);
+	tty_putp(cursor_normal, 0);
```
DBOTW: Regardless of the mode, the cursor should set to "normal" again ...

calleerlandsson:
Hey @DBOTW, cool that you were able to do this in so few lines of code!

> [...] if we where to go down the
route of adding Vi-emulation my fear is that people would want more and
more key bindings supported from Vi; potentially ending up lots of code.

I agree with @mptre's sentiment and I'm also worried that this feature will increase the maintenance burden and make refactoring harder.

I really appreciate your work, @DBOTW, and @mptre for taking the time to review this PR but for the above reasons, I'll have to vote against merging it.

DBOTW:
Thanks, @calleerlandsson! This was only a "proof of concept" and I share your concerns (it will remain in my tree and kept in sync with current versions of `pick`) ... so this and issue https://github.com/calleerlandsson/pick/issues/146 can be closed ...

Issue #62, opened 2015-06-23
Handling empty input


mattix:
Should empty input be ignored or handled in some other way? Currently only CTRL+C will exit.

To reproduce:
echo -ne "" | pick

calleerlandsson:
@mattix Great question! Sadly, I don't have a good answer for you. Do you have any ideas on how to best handle this?

mattix:
Sure thing!

Maybe simplest solution which comes to mind would be adding:

```
if (!SLIST_EMPTY(choices)) {
	io_print_choice(
		ui_selected_choice(choices, query, use_alternate_screen),
		output_description);
}
```
to https://github.com/thoughtbot/pick/blob/master/src/main.c#L60

This way pick will do nothing and return nothing if input is empty.

mattix:
This way:
https://github.com/mattix/pick/blob/master/src/main.c

calleerlandsson:
That's a nice solution!

I wonder, though, if we should exit with an exit code indicating failure since we weren't able to produce any meaningful output? Maybe we could even print an error message on `stderr` saying pick requires input?

I guess it depends on the use case.

Do you have a specific use case for this behavior, @mattix?

@rosshadden, do you have any thoughts on this?

I think #58, might be sort of related. If we exit immediately unless we get input, the user won't be able to use the new feature in #61.

rosshadden:
I think it's best to maintain consistent behavior. If the list is empty, it's still a list, it just has zero items. While I understand the motivation, I do not consider zero choices an "error", especially not one worthy of a nonzero exit code. 

Just my two cents.

mattix:
I wouldn't print any error either. The only choice for empty list would be empty I guess.

calleerlandsson:
@rosshadden, @mattix All good points.

@mattix, what would the use-case for this be? I have personally not run into this issue before.

calleerlandsson:
@mattix do you have any ideas on what use-case this would support?

calleerlandsson:
I'm going to close this due to inactivity.

PR #262, opened 2017-11-04
Exit on empty set of choices

DBOTW:
```sh
echo -n "" | pick
```
should not come up with a selection ...

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -177,22 +177,23 @@ main(int argc, char *argv[])
 			err(1, NULL);
 	}
 
-	input = get_choices();
-	tty_init(1);
+	if ((input = get_choices()) != NULL) {
 
 #ifdef HAVE_PLEDGE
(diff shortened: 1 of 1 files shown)


DBOTW:
> See #62.

OK, I've not digged in the cementary of issues ... but see no decision in #62  ... close this also?

mptre:
See #62.

mptre:
I'm not in favor of the proposed behavior. Other interactive utilities
such as less/more does not exit in the case of absent input:

```sh
$ less </dev/null
```

DBOTW:
> Other interactive utilities such as less/more does not exit in the case of absent input

Can't confirm this for: `cat` and `more` on HP-UX 10.20, NetBSD 7.1 & Solaris 9, `cat` from GNU coreutils 8.23, `more` from  util-linux 2.25.2, `less` (358, 458 and 481) ...

PR #263, opened 2017-11-05
Add toggle sorting key binding

mptre:
I use pick to find Git commits, a lot. I generally make use of the `-S`
option since newer commits (lines near the beginning of input) is often
more relevant than older ones with a higher score. However, it would be
convenient to be able to toggle sorting on and off.

Unfortunately, this binding conflicts with [STOP](http://man.openbsd.org/termios#VSTOP), might not be a problem
but proposals on a different binding are welcome.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,6 +17,7 @@ TESTS=tests/key-alt-enter.t \
       tests/key-ctrl-e.t \
       tests/key-ctrl-k.t \
       tests/key-ctrl-l.t \
+      tests/key-ctrl-o.t \
       tests/key-ctrl-u.t \
       tests/key-ctrl-w.t \
       tests/key-del.t \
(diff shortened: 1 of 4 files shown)


DBOTW:
> Unfortunately, this binding conflicts with STOP, might not be a problem

Software flow control (XON/XOFF) will not work ...

>  but proposals on a different binding are welcome.

How about `Ctrl-O` (for "sOrt" / "Order")?

mptre:
Nice, since `IEXTEN` is disabled it won't conflict with [`DISCARD`](http://man.openbsd.org/termios#DISCARD), [`VDISCARD`](http://man.openbsd.org/termios#VDISCARD).

calleerlandsson:
Nice! 👍 This looks good to me!

PR #264, opened 2017-11-06
Only redraw the choices when necessary

mptre:
Some key bindings only affect the query and unconditionally redrawing
the choices is therefore a wast of time in such scenarios.

Originally proposed by Jenz Guenther in PR #254.


mptre:
Sorry for taking so long. See #264 for a slightly different approach,
reducing the necessary indent level a bit.

PR #265, opened 2017-11-08
Change exit codes to C standard macros

DBOTW:


calleerlandsson:
Hey @DBOTW, thanks for the PR!

In most cases, I think naming "magic numbers" is a great practice. In this specific case, though, my gut reaction is that the verbosity of `EXIT_SUCCESS` doesn't add much value over `0`.

It's hard to put my finger on exactly how this case differs from other cases. Maybe be it's because the constants are so ubiquitous?

Another case where I'd make the same call is in the case of HTTP status codes. To me, `404` is clearer than `NOT_FOUND`.

What are your thoughts, @mptre?

mptre:
> What are your thoughts, @mptre?

I agree. With some knowledge of the semantics of exit codes in Unix I
find the macros too noisy.

DBOTW:
The macros are too long: `SUCCESS` and `FAILURE` would IMHO be sufficient.
There's no accounting for taste ... so closing this without making a noise. ;-)

PR #267, opened 2017-12-28
Use posix_fadvise for sequential read

DBOTW:
... this should improve performance of `read` on some platforms ... not sure if it makes sense on `stdin`.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -1,10 +1,16 @@
 /* config.h.in.  Generated from configure.ac by autoheader.  */
 
+/* Define to 1 if you have the <fcntl.h> header file. */
+#undef HAVE_FCNTL_H
+
 /* Define if ncursesw is available */
 #undef HAVE_NCURSESW_H
 
 /* Define to 1 if you have the `pledge' function. */
 #undef HAVE_PLEDGE
+
+/* Define to 1 if you have the `posix_fadvise' function. */
+#undef HAVE_POSIX_FADVISE
 
 /* Define to 1 if you have the `reallocarray' function. */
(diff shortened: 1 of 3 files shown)


mptre:
Interesting, do you have any numbers on the actual performance gain? How
large must the input be in order to notice a difference?

DBOTW:
> How large must the input be in order to notice a difference?

Please don't expect too much: there will probably be no remarkable performance gain ...
Some tools from `coreutils` use it:
https://github.com/coreutils/coreutils/blob/master/src/cat.c#L689
https://github.com/coreutils/coreutils/blob/master/src/copy.c#L1225
https://github.com/coreutils/coreutils/blob/master/src/wc.c#L225-L227

mptre:
I'm a bit reluctant to adding non-portable code if there's no clear
performance or security gain. Thoughts @calleerlandsson?

DBOTW:
> to adding non-portable code

http://pubs.opengroup.org/onlinepubs/009695399/functions/posix_fadvise.html

... of course this doesn't mean anything ... ;-)

calleerlandsson:
Hey @DBOTW, I didn't know about `posix_fadvise` so thanks for teaching me about it!

Unless there's a clear benefit to adding this hinting, I'm also a reluctant of doing so since it makes the code noisier.

DBOTW:
Hi @calleerlandsson:

> Unless there's a clear benefit to adding this hinting

Depends on the OS: FreeBSD >= 10, Linux >= 2.5.60 and NetBSD >= 4 support it ... but I cannot confirm a clear benefit of the performance gain ... so closing this ... ;-)

PR #268, opened 2017-12-29
Return status of filtering choices

DBOTW:
...  there is an issue here with aborted filtering on a large set of choices: sometimes "outdated choices" remain on the screen if filtering is rapidly aborted (by holding the `BackSpace`-Key to erase the query) ...

Erasing the query by slowly pressing `BackSpace`:
slow

Erasing the query by holding `BackSpace`:
fast

Erasing the query by holding `BackSpace` (fix applied):
fix

PR #270, opened 2017-12-30
Ensure correctness while typing a query fast

mptre:
Typing a query fast enough to trigger the poll check in filter_choices()
could yield a wrong set of choices. Since filter_choices() is
interrupted, all choices have yet not been examined and potential
matches could still be left unexamined. Calling print_choices() in this
state causes the unexamined choices to never be reconsidered as
potential matches since they will have a score = 0. The solution is to
never call print_choices() if filter_choices() was interrupted.

Problem reported and partial solution provided by Jenz Guenther in
GitHub issue #268.

### DIFF
--- a/pick.c
+++ b/pick.c
@@ -71,7 +71,7 @@ struct choice {
 static int			 choicecmp(const void *, const void *);
 static void			 delete_between(char *, size_t, size_t, size_t);
 static char			*eager_strpbrk(const char *, const char *);
-static void			 filter_choices(size_t);
+static int			 filter_choices(size_t);
 static char			*get_choices(void);
 static enum key			 get_key(const char **);
 static void			 handle_sigwinch(int);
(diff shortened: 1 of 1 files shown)


DBOTW:
OK, but by pressing another "special key" (`Home` or `End`) directly after typing the query fast, the screen remains blank here ... IMHO should `dofilter` only be reseted if `filter_choices` returns success ...

The reason for
```C
case CTRL_A:
        cursor_position = 0;
        if (!dofilter)
                dochoices = 0;
        break;
```
in https://github.com/calleerlandsson/pick/pull/268/commits/59ddc0f093d6e7848164653e162970845f687f35 is to avoid a blank screen (with no choices) for `Ctrl-A`, `Ctrl-E`, `Left` and `Right` ...

mptre:
Ah, good catch. Just updated #270 with a potential fix. I this works,
the `dochoices` logic could be simplified since `dochoices = 1` should
only be implied if `filter_choices()` returns `1`. Does that make sense?

DBOTW:
Great! Works perfect! Closing https://github.com/calleerlandsson/pick/pull/268 in favor for this ...

mptre:
Nice! Please try the latest commit which includes a simplification of
the `dochoices` logic.

DBOTW:
That's it! Time for version 2.0.1 of `pick`! :-)

mptre:
Yes! I'm planning on releasing 2.0.1 next year (that's two days from now :trollface:)

PR #271, opened 2017-12-31
Add emulation for pledge with seccomp

DBOTW:
... for https://github.com/calleerlandsson/pick/issues/269 ... it's horrible but compiles and seems to work ... needs some more testing ...

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,7 @@ AM_CFLAGS=-Wall -Werror -Wextra
 AM_CPPFLAGS=-D_GNU_SOURCE
 
 bin_PROGRAMS=pick
-pick_SOURCES=pick.c compat-reallocarray.c compat-strtonum.c compat.h
+pick_SOURCES=pick.c compat-pledge.c compat-reallocarray.c compat-strtonum.c compat.h
 pick_CPPFLAGS=$(AM_CPPFLAGS) $(NCURSES_CFLAGS)
 pick_LDADD=$(NCURSES_LIBS)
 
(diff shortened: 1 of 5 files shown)


mptre:
Thanks a great start @DBOTW! See #273, I wonder if we instead of
mimicking pledge on different OSs should come up with a different
abstraction and reason about the privilegies related to the state of
pick. Would really appreciated your thoughts.

/cc @DBOTW @mike-burns  @calleerlandsson

DBOTW:
Thanks @mptre and @mike-burns, but this was just a quick hack based on
https://github.com/aggsol/linux-pledge/ and modified for the needs of `pick` ...

`strace` was used to determine the `syscalls` that terminate the execution of
`pick` and these were added to the "rules" (marked with `/* Necessary */`).

This is probably not the right approach when it comes to "security": by adding
more exceptions to a packet filter, after a certain time it's useless anyway ...

Even worse: I'm using `seccomp` for the first time, so chances are high that the
added exceptions "undermine" the whole "security" (`RULE(prctl);`) ...

So I'm concerned about the patch (that said I'm also not qualified for this) ...

DBOTW:
OK, not that bad: `RULE(prctl);` is necessary because `pledge` is called a second time (of course without `prctl` no rules can be defined) ...

I have removed `rpath wpath cpath` from the first call and it still works here, so with https://github.com/calleerlandsson/pick/pull/273 there is better way to "fine tune" privilegies depending on the OS ...

DBOTW:
Closing this in favor of https://github.com/calleerlandsson/pick/pull/274 ...

PR #272, opened 2018-01-01
Remove Werror compiler flag

mptre:
It's generally considered bad practice to ship software with this flag.
Issue revealed while updating the port on [OpenBSD].

[OpenBSD]: https://marc.info/?l=openbsd-ports&m=151482370923381&w=2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,10 @@ matrix:
   include:
     - os: linux
       compiler: clang
-      env: BUILD_TYPE=default CFLAGS='-fsanitize=address -fsanitize=undefined'
+      env: BUILD_TYPE=default CFLAGS='-fsanitize=address -fsanitize=undefined -Werror'
     - os: linux
       compiler: clang-5.0
-      env: BUILD_TYPE=default CFLAGS='-fsanitize=address -fsanitize=undefined'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,10 +8,10 @@ matrix:
   include:
     - os: linux
       compiler: clang
-      env: BUILD_TYPE=default CFLAGS='-fsanitize=address -fsanitize=undefined'
+      env: BUILD_TYPE=default CFLAGS='-fsanitize=address -fsanitize=undefined -Werror'
     - os: linux
       compiler: clang-5.0
-      env: BUILD_TYPE=default CFLAGS='-fsanitize=address -fsanitize=undefined'
+      env: BUILD_TYPE=default CFLAGS=' -fsanitize=address -fsanitize=undefined -Werror'
```
calleerlandsson: It seems like a space snuck in just before `-fsanitize`.

mptre: Nice catch!

mptre:
This was a new low on Travis:

```
configure:3030: ./conftest
==LeakSanitizer has encountered a fatal error.
```

mptre:
This seems to be the cause travis-ci/travis-ci#9033.

Issue #269, opened 2017-12-30
Sanbox support for more OS


mptre:
It would be interesting to investigate adding support for sandboxing on
operating systems other than OpenBSD where [pledge] is available.
There's seccomp on Linux and Capsicum on FreeBSD. I wouldn't worry about
the portability at this stage, just the code required to get it working
would be a great start.

[pledge]: http://man.openbsd.org/pledge

PR #273, opened 2018-01-01
Sandbox abstraction proposal

mptre:
Related to GitHub issue #269 and #271.


calleerlandsson:
This is really cool! Thanks for the great work @DBOTW and @mptre. Should #274 be merged into the branch for this PR?

It's not clear to me why the build is failing or why it's not outputting "disallowed syscall" messages.

DBOTW:
> It's not clear to me why the build is failing or why it's not outputting "disallowed syscall" messages

Seems that `travis` needs more allowed syscalls (as errorcode `159` indicates, `pick` was terminated by `seccomp`), but it's hard to figure out: local builds (i386/amd64) don't need `futex` or `getgid` here ...

`strace` is disabled by default in `docker` (`PTRACE_TRACEME doesn't work: Operation not permitted`), so it's "trial & error" ... adding the "needed" syscalls makes `travis`  happy but is probably not necessary on "bare hardware", so there's no gain of security here ... don't know how to solve this conflict ... :-(

mptre:
> It's not clear to me why the build is failing or why it's not outputting "disallowed syscall" messages.

I've been trying to figure out why the disallowed message doesn't show
up on Travis. Even strace:ing pick shows that the write call succeeds. I
got a bit reluctant to shipping support for seccomp after reading
[this]. The plan was to make `make check || cat test-suite.log` output
the missing syscall to ease debugging on platforms experiencing
failures.

[this]: https://github.com/kristapsdz/acme-client-portable/blob/master/Linux-seccomp.md

DBOTW:
> I got a bit reluctant to shipping support for seccomp

`seccomp` should be disabled by default but can be used with `--enable-seccomp`  ...
(latest https://github.com/calleerlandsson/pick/pull/274 has been modified accordingly)

DBOTW:
> I've been trying to figure out why the disallowed message doesn't show up

Redirecting `stderr` to a file instead of `2>&1` in `pick-test.sh`?

DBOTW:
Quoting https://github.com/google/sanitizers/issues/777#issuecomment-284220988:

> Our general suggestion is to disable any sandboxes in sanitizer builds.

... I'm doing so in latest https://github.com/calleerlandsson/pick/pull/274 ...

PR #276, opened 2018-02-11
Remove keypad transmit mode and hardcode cursor keys

DBOTW:
... for https://github.com/calleerlandsson/pick/issues/275 ...

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -36,7 +36,6 @@ TESTS=tests/key-alt-enter.t \
       tests/misc-realloc.t \
       tests/opt-d.t \
       tests/opt-h.t \
-      tests/opt-k.t \
       tests/opt-o.t \
       tests/opt-q.t \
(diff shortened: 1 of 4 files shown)


mptre:
This breaks the cursor keys in xterm for me since it now emits `\e[`
prefix as opposed of `\eO`. Instead of dropping transmit mode, can't we
just hardcode cursor keys with the `\eO` prefix instead? That would
requires less sync with pick.vim and no major bump.

mptre:
If this is a viable approach, I'd like to see tests for added cursor keys.

DBOTW:
> If this is a viable approach, I'd like to see tests for added cursor keys.

Of course (see https://github.com/calleerlandsson/pick/pull/277) ... fortunately the tests are already there ...

mptre:
@DBOTW you're right!

PR #280, opened 2018-03-24
Trim down README

mptre:
- Merge the initial paragraph with usage, now that we can refer to manual

- Limitations of calling pick from within Vim is better of described in the
  README of pick.vim

/cc @calleerlandsson @teoljungberg @mike-burns

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Copyright (c) 2017 Calle Erlandsson, Anton Lindqvist & thoughtbot.
 
 [crux]: https://github.com/6c37/crux-ports
 [current]: https://github.com/calleerlandsson/pick/blob/master/CONTRIBUTING.md
-[debian]: https://packages.debian.org/stretch/pick
+[debian]: https://packages.debian.org/stable/misc/pick
 [gentoo]: https://packages.gentoo.org/packages/sys-apps/pick
 [pick-vim]: https://github.com/calleerlandsson/pick.vim
 [pick]: https://calleerlandsson.github.io/pick/
--- a/README.md
+++ b/README.md
@@ -1,11 +1,23 @@
-# Pick
+# pick
 
 ![pick(1) usage](screencast.gif)
 
-The
 [pick(1)][pick]
-utility allows users to choose one option from a set of choices using an
-interface with fuzzy search functionality.
+reads a list of choices from `stdin` and outputs the selected choice to
+`stdout`.
+Therefore it is easily used both in pipelines and subshells:
+
+```sh
+# Select a file in the current directory to open using xdg-open(1):
+$ find . -type f | pick | xargs xdg-open
+# Select a command from the history to execute:
+$ eval $(fc -ln 1 | pick)
+```
+
+[pick(1)][pick] can also be used from within Vim,
+see the
+[pick.vim][pick-vim]
+plugin.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -111,6 +93,8 @@ Copyright (c) 2017 Calle Erlandsson, Anton Lindqvist & thoughtbot.
 [current]: https://github.com/calleerlandsson/pick/blob/master/CONTRIBUTING.md
 [debian]: https://packages.debian.org/stretch/pick
 [gentoo]: https://packages.gentoo.org/packages/sys-apps/pick
+[pick-examples]: https://mptre.github.io/pick.1#EXAMPLES
```
teoljungberg: Should this link point to the original repository?

mptre: Not sure I follow. This is a reference to the HTML rendered manual.

teoljungberg: Right, I'm thinking it may cause confusion to users being redirected to another GitHub users website rather than the ones they started from. In this case Calle's project to your GitHub site.

It may be semantics, and doesn't really matter.

mptre: Ah, we could stick it in the `gh-pages` branch for pick instead.

mptre: Done

README.md
```diff
@@ -111,7 +95,9 @@ Copyright (c) 2017 Calle Erlandsson, Anton Lindqvist & thoughtbot.
 [current]: https://github.com/calleerlandsson/pick/blob/master/CONTRIBUTING.md
 [debian]: https://packages.debian.org/stretch/pick
```
DBOTW: This should point to: https://packages.debian.org/stable/misc/pick

mptre: Fixed in 65db1c9.

mike-burns:
I like it, but I'd prefer to still see at least one usage example in the README to help non-sighted readers understand what it does at a glance. (There's the GIF at the top but that won't help people using a screenreader.)

Just this amount is what I request:

> pick(1) reads a list of choices on `stdin` and outputs the selected choice on
> `stdout`. Therefore it is easily used both in pipelines and subshells:
> 
> ```sh
> git ls-files | pick | xargs less # Select a file in the current git repository to view in less
> cd "$(find . -type d | pick)"    # Select a directory to cd into
> eval $(fc -ln 1 | pick)          # Select a command from the history to execute
> ```

While here, I like tiny usage snippets like that to go above installation, as a sales pitch to the new potential user.

What do you think?

calleerlandsson:
I agree with @mike-burns that a tiny bit of usage can be helpful in showing what the program does rather than explaining it. Otherwise, I think this looks great!  🎉

Do I have to do something to make the https://calleerlandsson.github.io/pick/ links work?

mptre:
Thanks for the feedback!

> What do you think?

Makes perfect sense. I made one modification and used the examples from
the manual instead.

> Do I have to do something to make the https://calleerlandsson.github.io/pick/ links work?

I thought it was a cache issue at first. Check `Settings > Options >
GitHub Pages > Source` and ensure `gh-pages` is selected.

mike-burns:
LGTM!

calleerlandsson:
> I thought it was a cache issue at first. Check `Settings > Options >
GitHub Pages > Source` and ensure `gh-pages` is selected. 

`gh-pages` is the selected branch. According to the docs, this should be enough. Maybe a new commit to the branch will trigger a build?

This PR looks good to merge!

mptre:
> gh-pages is the selected branch. According to the docs, this should be enough. Maybe a new commit to the branch will trigger a build?

Still no luck...

calleerlandsson:
I turned it off and on again and now it seems to work 🤷‍♂️...

PR #282, opened 2018-04-02
Limit usage of malloc.conf options to OpenBSD

mptre:
Turns out the S option has a different meaning on {Free,Net}BSD and the R option
is not supported.

/cc @DBOTW and @mike-burns for autoconf inspection


configure.ac
```diff
@@ -15,5 +15,19 @@ AC_SEARCH_LIBS([setupterm], [curses], [], [
     )]
   )
 ])
+AC_DEFUN([AC_MALLOC_OPTIONS], [
+  AC_MSG_CHECKING([for malloc hardening options])
+  case "`uname -s`" in
```
mike-burns: If you [`AC_CANONICAL_HOST`](https://www.gnu.org/software/autoconf/manual/autoconf-2.62/html_node/Canonicalizing.html) then you can use `$host_os` in here instead.

mptre:
Thanks @mike-burns! Updated diff with your proposal and significant
rework:

- Make sure to always run `AC_SUBST([MALLOC_OPTIONS], ...)`

- Only set `MALLOC_OPTIONS` when invoking pick so it won't affect any
  other intermediate processes along the process tree leading down to pick

@DBOTW could you confirm that it's working as expected on NetBSD?

PR #285, opened 2018-04-15
Event driven design

mptre:
This is probably my last remaining annoyance with pick, the fact that it does a
blocking read of stdin until EOF prior to entering its main loop. If stdin is
slow, this will take time; like running git-log in a huge repository.

The new approach uses poll(2) and can therefore read from stdin as new data
arrives and simultaneously handle user commands. Some notes about the changes:

- get_choices() is made re-entrant. It still allocates a single block of memory
  for the stdin data. Each choice stores an indirect pointer to this piece of
  memory since it could be realloacted invalidating any previously stored
  pointer. This approach is more complex than storing the string in each choice
  but much faster due to less allocations and strdup(3):ing.

- tty_getc() no longer uses the FILE API from libc since it's buffered. Meaning,
  it could potentially read more input stored in its internal buffer during a
  single call to getc(3). If all that data is not consumed during the call to
  get_key(), poll will never flag the underlying file descriptor as non-block
  readable again since the data is already consumed. Issue exhibited by the test
  suite.

To see it in action:

```sh
$ while :; do sleep 1; echo bar; done | pick -q bar
```

Testing and feedback would be much appreciated.

/cc @calleerlandsson @mike-burns @DBOTW @teoljungberg


teoljungberg:
I'm really excited for a feature such as this!

I'm on macOS, which really might be the culprit, and I can't really seem to get this to run properly. 
Your example isn't returning me anything, and seems to reset my TTY. After re-running a command with `pick(1)` I'm prompted with:

```
% git ls-files | ./pick -q pick
pick: 3: bad file descriptor
```

Is there a difference between how file descriptors are on macOS and on OpenBSD (your OS of choice) that messes with this?

teoljungberg:
Just as I commented I saw the update in 423c87dd2a7c5ecb180b1b33f23f054be0f9c73c, and I have the same experience with that patch too.

mptre:
> Just as I commented I saw the update in 423c87dd2a7c5ecb180b1b33f23f054be0f9c73c, and I have the same experience with that patch too.

@teoljungberg bummer, I'm out of ideas on how-to solve this on macOS.
Anyone aware how others have tackled this problem?

mptre:
> This is awesome and https://github.com/mptre/pick/issues/22 can finally be closed :+1: ... but on some use cases the old method is faster:

@DBOTW try the latest commit on this branch. Does it improve the
performance? It's quite messy right now but the idea is to prepend the
newly read choices limiting the number of choices that needs to examined
during search.

mptre:
> ... but pasting a query with middle-click does not work if no key is pressed before ...

Try the latest commit.

DBOTW:
It works only with this change:
```diff
diff --git a/pick.c b/pick.c
index fd64bc5..d0b7d82 100644
--- a/pick.c
+++ b/pick.c
@@ -535,7 +535,7 @@ selected_choice(void)
                        cursor_position += length;
                        query_length += length;
                        query[query_length] = '\0';
-                       dofilter = 1;
+                       dofilter = choices_reset = 1;
                        break;
                case UNKNOWN:
                        break;
```

teoljungberg:
> I'm out of ideas on how-to solve this on macOS.
Anyone aware how others have tackled this problem?

@mptre That's sadly more than I can help you with. Maybe @calleerlandsson has a trick up his sleeve?

mptre:
@teoljungberg try running the following program, let i run for a second
then hit enter:

```sh
$ cc -x c -o tty - <<EOF && ./tty
#include <err.h>
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>

int
main(void)
{
	ssize_t n;
	int fd, flags;
	unsigned char c;

	fd = open("/dev/tty", O_RDONLY);
	if (fd == -1)
		err(1, "open");
	flags = fcntl(fd, F_GETFL, 0);
	if (flags == -1)
		err(1, "fcntl: F_GETFL");
	if (fcntl(fd, F_SETFL, flags | O_NONBLOCK) == -1)
		err(1, "fcntl: F_SETFL");
	for (;;) {
		n = read(fd, &c, 1);
		fprintf(stderr, "%s: n=%ld\n", __func__, n);
		if (n == -1 && errno != EAGAIN)
			err(1, "read");
		if (n == 1)
			break;
		sleep(1);
	}
	close(fd);
	fprintf(stderr, "%#x\n", c);
	return 0;
}
EOF
```

My latest idea is as follows: while stdin has not yet reached EOF, poll
from stdin with a fairly short timeout (100ms), then do a non-blocking
read from the tty. Once stdin has reached EOF, we can resort to blocking
read of the tty.

teoljungberg:
@mptre That seems about right:

```
main: n=-1

main: n=1
0xa
```

DBOTW:
@mptre Works perfect now! :+1:

mptre:
@teoljungberg please try the latest commit.

teoljungberg:
@mptre It compiles!

I did a crazy thing and tried `% find $HOME/src -type f | pick -q pick` - and the performance was less than stellar. But it's difficult to debug if it's due to this PR as I would never dream of running it before this event driven change.

```
% find $HOME/src -type f | wc -l
  217274
```

But - since we are now tentatively handling more files. Are there performance gains we can gain somehow?

teoljungberg:
It looks like the polling is still occurring after selection has been "selected".

```
% find $HOME -type l | pick -q ctags
/Users/teo/.vim/pack/bundle/opt/rust.vim/.git/hooks/ctags







ls
~ %






ls
~ %





ls
~ %




ls
~ %



ls
~ %


ls
~ %

ls
~ %
ls
~ % ls
```

DBOTW:
Tested latest commit: blazing fast! :1st_place_medal:
I suggest the following change that allows scrolling while the list is filled:
```diff
diff --git a/pick.c b/pick.c
index 1af8abb..d32c589 100644
--- a/pick.c
+++ b/pick.c
@@ -526,6 +526,7 @@ selected_choice(void)
                        cursor_position += length;
                        query_length += length;
                        query[query_length] = '\0';
+                       selection = yscroll = 0;
                        dofilter = 1;
                        break;
                case UNKNOWN:
@@ -533,12 +534,14 @@ selected_choice(void)
                }
 
 render:
-               if (choices_reset)
+               if (choices_reset) {
                        choices_count = choices.length;
+                       selection = yscroll = 0;
+               }
                choices_reset = 0;
                if (dofilter) {
                        if ((dochoices = filter_choices(choices_count)))
-                               dofilter = selection = yscroll = 0;
+                               dofilter = 0;
                }
 
                tty_putp(cursor_invisible, 0);
```

mptre:
> Tested latest commit: blazing fast! :1st_place_medal:

Glad to hear!

> I suggest the following change that allows scrolling while the list is filled:

Also tried to fix this. There's one problem with your diff:

```sh
$ (echo a; echo aa;) | pick) # type: a DOWN a
```

... the selection is now out of bounds.

mptre:
> It looks like the polling is still occurring after selection has been "selected".

@teoljungberg not sure I follow, does it leave your terminal in a weird state?

teoljungberg:
@mptre it does, and only if I select something while the list to pick is still being populated from the stream.

Worth noting, this still happens with tools such as `fzf`.

teoljungberg:
Now that I compared it to the speed of `fzf` - it's noticeably slower to type in `pick` for a list of the same size.

mptre:
Just pushed another performance improvement. Here's some number of the actual
performance on my machine:

```sh
$ find /usr/src -type f | wc -l
85078
$ find /usr/src -type f | env LINES=10 ./pick -Xq realloc
master branch:                       2.3s
poll branch:                         8.0s
poll branch with latest improvement: 2.4s
```

The performance looks similar to the current version of pick. This is not the
best test to demonstrate the advantage of this diff since stdin is producing
input at a high pace. But, the real advantage is that the UI is more responsive
and gives the illusion of running fast since the it's updated continuously. The
real gain comes when stdin is slower.

It would be interesting if anyone else could perform similar benchmarks using a
large piece of input with rougly the same amount of matches (2442) as I did.
Below is two patches to give an idea on how to get the numbers for both the poll
and master branch:


### poll

```diff
diff --git a/pick.c b/pick.c
index f128349..5fe50e1 100644
--- a/pick.c
+++ b/pick.c
@@ -335,6 +335,9 @@ selected_choice(void)
 	int		 dochoices = 0;
 	int		 dofilter = 1;
 
+	struct timespec t0, t1, t2;
+	clock_gettime(CLOCK_MONOTONIC, &t0);
+
 	cursor_position = query_length;
 
 	/* No timeout on first call to fdwait in order to render the UI fast. */
@@ -370,6 +373,9 @@ selected_choice(void)
 				}
 			} else {
 				fd = -1; /* EOF */
+				clock_gettime(CLOCK_MONOTONIC, &t1);
+				timespecsub(&t1, &t0, &t2);
+				fprintf(stderr, "%lld.%ld\n", t2.tv_sec, t2.tv_nsec);
 			}
 		}
 		if ((ready & FD_TTY_READY) == 0)
```

### master

```diff
diff --git a/pick.c b/pick.c
index 4ba17de..56550fd 100644
--- a/pick.c
+++ b/pick.c
@@ -33,6 +33,8 @@
 		errx(1, #capability ": unknown terminfo capability");	\
 } while (0)
 
+struct timespec t0, t1, t2;
+
 enum key {
 	UNKNOWN,
 	ALT_ENTER,
@@ -178,6 +180,7 @@ main(int argc, char *argv[])
 			err(1, NULL);
 	}
 
+	clock_gettime(CLOCK_MONOTONIC, &t0);
 	input = get_choices();
 	tty_init(1);
 
@@ -354,6 +357,9 @@ selected_choice(void)
 		tty_putp(cursor_normal, 0);
 		fflush(tty_out);
 
+		clock_gettime(CLOCK_MONOTONIC, &t1);
+		timespecsub(&t1, &t0, &t2);
+		fprintf(stderr, "%lld.%ld\n", t2.tv_sec, t2.tv_nsec);
 		switch (get_key(&buf)) {
 		case ENTER:
 			if (choices_count > 0)
```

DBOTW:
```sh
$ (echo a; echo aa;) | pick # type: a DOWN a
```
... with latest commit and the diff:
scroll

teoljungberg:
@mptre I'd be glad to try it out, but after I applied your diff by hand (it would have been neat to have a patch to apply, for future reference) I got the following error:

```
% make
/Applications/Xcode.app/Contents/Developer/usr/bin/make  all-am
gcc -DHAVE_CONFIG_H -I.  -D_GNU_SOURCE   -Wall -Wextra -g -O2 -MT pick-pick.o -MD -MP -MF .deps/pick-pick.Tpo -c -o pick-pick.o `test -f 'pick.c' || echo './'`pick.c
pick.c:377:5: warning: implicit declaration of function 'timespecsub' is invalid in C99 [-Wimplicit-function-declaration]
                                timespecsub(&t1, &t0, &t2);
                                ^
pick.c:378:35: warning: format specifies type 'long long' but the argument has type '__darwin_time_t' (aka 'long') [-Wformat]
                                fprintf(stderr, "%lld.%ld\n", t2.tv_sec, t2.tv_nsec);
                                                 ~~~~         ^~~~~~~~~
                                                 %ld
2 warnings generated.
mv -f .deps/pick-pick.Tpo .deps/pick-pick.Po
gcc -Wall -Wextra -g -O2   -o pick pick-compat-fdwait.o pick-compat-reallocarray.o pick-compat-strtonum.o pick-pick.o  -lcurses
Undefined symbols for architecture x86_64:
  "_timespecsub", referenced from:
      _main in pick-pick.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [pick] Error 1
make: *** [all] Error 2
```

The latter is easy enough to fix, change `%lld` to %ld`. But the second I'm not sure about how to tackle.

mptre:
@teoljungberg https://github.com/openbsd/src/blob/master/sys/sys/time.h#L126-L134

DBOTW:
```sh
% find . -name "*" | wc -l
74447
% find . -name "*" | grep "mod" | wc -l
1636
% find . -name "*" | env LINES=10 pick -Xq mod

latest master: 1.426684576
latest poll:   1.553520823
```

teoljungberg:
Sorry @mptre - this goes beyond my level of C. I'll happily run benchmarks to help out, but I don't know what to do here.

mptre:
Thanks @DBOTW and @teoljungberg. So the performance looks the same on macOS? If that's true, I think this approach is worth pursuing. I will try to get hold of a macOS machine in order to investigate abort during poll problem.

I would really appreciate if you guys could keep running this branch on a daily basis to shake out potential bugs.

DBOTW:
Running 10 times:
```sh
% find $HOME/linux-master -type f | env LINES=10 ./pick -Xq define
```
master:
```sh
1.26849022
1.9111261
1.33330356
1.1619533
1.9843511
1.6209546
0.994171390
1.7731074
1.35363015
1.11947793
Average:
1.452056575
```
poll:
```sh
1.130832072
1.113641899
1.129963565
1.121772860
1.153440767
1.110494182
1.93199549
1.112995187
1.108392331
1.98562282
Average:
1.289915117
```

teoljungberg:
I also ran the benchmarks 10 times to give a proper average:

`poll`
```
0.465581000
0.452371000
0.445096000
0.482430000
0.450860000
0.470767000
0.449865000
0.466116000
0.444745000
0.449640000

Average:
0.457747
```

`master`
```
0.474734000
0.462568000
0.465852000
0.458278000
0.462156000
0.463605000
0.462227000
0.467203000
0.457667000
0.464166000

Average:
0.463846
```

mptre:
Reworked the code a bit, but it should not cause any functional difference. Please test!

DBOTW:
`gcc` (version `4.9.2` and older) shows:
```sh
pickpick.c: In function ‘main’:
pick.c:332:11: warning: ‘choices_offset’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  size_t   choices_offset, cursor_position, i, j, length, nfds,
           ^
```
so:
```diff
diff --git a/pick.c b/pick.c
index 9aa9a81..ce42794 100644
--- a/pick.c
+++ b/pick.c
@@ -329,9 +329,9 @@ selected_choice(void)
        struct pollfd    fds[2];
        const char      *buf;
        ssize_t          insert;
-       size_t           choices_offset, cursor_position, i, j, length, nfds,
-                        xscroll;
+       size_t           cursor_position, i, j, length, nfds, xscroll;
        size_t           choices_count = 0;
+       size_t           choices_offset = 0;
        size_t           selection = 0;
        size_t           yscroll = 0;
        int              dokey, doread, nready, timo;
```

DBOTW:
@mptre Sorry, can't repoduce the issue with https://github.com/mptre/pick/pull/285#issuecomment-381706116 as https://github.com/mptre/pick/pull/285#issuecomment-381841895 shows ...
the diff is not "perfect", but resetting the selection while scrolling is annyoing ... ;-)

DBOTW:
@mptre No warnings ... but the selection is still reset when new entries stream in that match the query.

DBOTW:
@mptre better :smiley: ... but by holding down `End` or `Page Down` after typing the query fast, some "phantom choices" appear ... the following diff fixes this:
```diff
diff --git a/pick.c b/pick.c
index db4f407..b981825 100644
--- a/pick.c
+++ b/pick.c
@@ -623,7 +623,7 @@ filter_choices(size_t offset, size_t *nchoices)
                if (c->score > 0 || query_length == 0)
                        n++;
 
-               if (i > 0 && i % 50 == 0) {
+               if (n > 0 && n % 50 == 0) {
                        pfd.fd = fileno(tty_in);
                        pfd.events = POLLIN;
                        if ((nready = xpoll(&pfd, 1, 0)) == -1)
```

DBOTW:
@mptre Tested and looks good ... :smiley:

>  one problem at a time

OK, here's the next one: by holding down a key `filter_choices` will be aborted and the choices are not filtered anymore, because the key is not passed to `get_key` ...

```diff
diff --git a/pick.c b/pick.c
index db4f407..b2ed42f 100644
--- a/pick.c
+++ b/pick.c
@@ -393,6 +393,7 @@ selected_choice(void)
                if (!dokey)
                        goto render;
 
+getkey:
                switch (get_key(&buf)) {
                case ENTER:
                        if (choices_count > 0)
@@ -561,6 +562,8 @@ render:
                            &choices_count);
                        if (dochoices)
                                dofilter = 0;
+                       else
+                               goto getkey;
                }
 
                tty_putp(cursor_invisible, 0);
```
Of course this is not a nice solution as it blocks the GUI until the key is released ...

teoljungberg:
I've been running this since the patch was announced, without anything major breaking.

Is there any final testing / tweaking we want to do?

DBOTW:
> I've been running this since the patch was announced, without anything major breaking.

... same here ...

> Is there any final testing / tweaking we want to do?

Good question. The requested code reviews are now pending since 3 months ...

mptre:
> > I've been running this since the patch was announced, without anything major breaking.

> ... same here ...

same here

I'm experiencing UI-flickering while stdin is being read that I'd like to address before getting this in.

DBOTW:
> I'm experiencing UI-flickering while stdin is being read that I'd like to address before getting this in.

This was IMHO caused by calling `fflush(tty_out)` too rapidly: the terminal needs some time to redraw the screen, but was interrupted by the next `fflush(tty_out)`. After various approaches with no satisfying results, it seems to me that the flickering can only be reduced, but not completely avoided ...

The simplest way is to add a small delay after `fflush(tty_out)` to wait for screen redraw:
```diff
diff --git a/pick.c b/pick.c
index db4f407..8b797c9 100644
--- a/pick.c
+++ b/pick.c
@@ -589,6 +589,8 @@ render:
                        tty_putp(tty_parm1(parm_right_cursor, j), 1);
                tty_putp(cursor_normal, 0);
                fflush(tty_out);
+               if (doread)
+                       usleep(10000);  /* Try to reduce screen flickering */
        }
 }

```

mptre:
> The simplest way is to add a small delay after fflush(tty_out) to wait for the terminal to redraw:

I'm thinking instead, we could avoid redrawing the interface when nothing changed. Like when new choices arrive on stdin but they don't affect the choices currently visible.

DBOTW:
> I'm thinking instead, we could avoid redrawing the interface when nothing changed

Right ... and this is the "core problem": `dochoices` will only be set to `0` when the filtering is aborted by pressing a key ... otherwise it remains at `1` and so the screen is redrawn every time (even if you only move the cursor in the query line: see https://github.com/mptre/pick/pull/264) ... the following diff should solve this:
```diff
diff --git a/pick.c b/pick.c
index db4f407..af46c40 100644
--- a/pick.c
+++ b/pick.c
@@ -335,7 +335,7 @@ selected_choice(void)
        size_t           yscroll = 0;
        int              dokey, doread, nready, timo;
        int              choices_reset = 1;
-       int              dochoices = 0;
+       int              dochoices = 1;
        int              dofilter = 1;
 
        cursor_position = query_length;
@@ -484,6 +484,7 @@ selected_choice(void)
                        break;
                case LINE_DOWN:
                        if (selection < choices_count - 1) {
+                               dochoices = 1;
                                selection++;
                                if (selection - yscroll == choices_lines)
                                        yscroll++;
@@ -491,6 +492,7 @@ selected_choice(void)
                        break;
                case LINE_UP:
                        if (selection > 0) {
+                               dochoices = 1;
                                selection--;
                                if (yscroll > selection)
                                        yscroll--;
@@ -511,19 +513,24 @@ selected_choice(void)
                                yscroll = selection += choices_lines;
                        else
                                selection = choices_count - 1;
+                       dochoices = 1;
                        break;
                case PAGE_UP:
                        if (selection > choices_lines)
                                yscroll = selection -= choices_lines;
                        else
                                yscroll = selection = 0;
+                       dochoices = 1;
                        break;
                case END:
-                       if (choices_count > 0)
+                       if (choices_count > 0) {
                                selection = choices_count - 1;
+                               dochoices = 1;
+                       }
                        break;
                case HOME:
                        yscroll = selection = 0;
+                       dochoices = 1;
                        break;
                case PRINTABLE:
                        if (query_length == 0)
@@ -556,7 +563,7 @@ render:
                if (choices_reset)
                        choices_count = choices.length;
                choices_reset = 0;
-               if (dofilter) {
+               if (dofilter && choices_count > 0) {
                        dochoices = filter_choices(choices_offset,
                            &choices_count);
                        if (dochoices)
@@ -576,6 +583,7 @@ render:
                        if (selection - yscroll >= choices_lines)
                                yscroll = selection - choices_lines + 1;
                        print_choices(yscroll, choices_count, selection);
+                       dochoices = 0;
                }
                tty_putp(carriage_return, 1);   /* move cursor to first column */
                for (i = j = 0; i < cursor_position; j++)
```

DBOTW:
Tested with:
```sh
% find $HOME/linux-master -type f | pick
```
Number of redraws WITHOUT the above patch:
`195`
Number of redraws  WITH the above patch:
 `1`

... no UI-flickering or other unnecessary redraws ...

/cc @mptre

DBOTW:
Running the patched version 10 times:
```sh
% find $HOME/linux-master -type f | env LINES=10 ./pick -Xq define
```
gives:
```
0.825561067
0.824530031
0.800953288
0.927038460
0.808921425
0.808347666
0.856313665
0.802832432
0.801190908
0.790501239
```
Average:
```
0.824619018
```
... compared to https://github.com/mptre/pick/pull/285#issuecomment-382505264 it's over `55%` faster ...

DBOTW:
@mptre Great! Tested your diff with
```sh
% find $HOME/linux-master -type f | env LINES=10 ./pick -Xq define
```
the number of redraws drops from `33` to `16` ...

```diff
diff --git a/pick.c b/pick.c
index 94a7a9f..cd7ab9c 100644
--- a/pick.c
+++ b/pick.c
@@ -573,7 +573,7 @@ render:
                if (choices_reset)
                        choices_count = choices.length;
                choices_reset = 0;
-               if (dofilter) {
+               if (dofilter && !dochoices) {
                        dochoices = filter_choices(choices_offset,
                            &choices_count);
                        if (dochoices)
```
or scrolling through the choices is not possible with the above example ...

DBOTW:
@mptre slightly modified diff:
- Removed the `didkey`-stuff (to be honest, I do not understand the reason ;-)
- Moved `realloc` of `choices_order` to `tty_size` to respect new screen dimensions
```diff
diff --git a/pick.c b/pick.c
index db4f407..ad15824 100644
--- a/pick.c
+++ b/pick.c
@@ -111,6 +111,7 @@ static struct {
 }                               choices;
 static FILE                    *tty_in, *tty_out;
 static char                    *query;
+static size_t                  *choices_order;
 static size_t                   query_length, query_size;
 static volatile sig_atomic_t    gotsigwinch;
 static unsigned int             choices_lines, tty_columns, tty_lines;
@@ -204,6 +205,7 @@ main(int argc, char *argv[])
                free(*choices.v[0].string);
        free(choices.v);
        free(query);
+       free(choices_order);
 
        return rc;
 }
@@ -335,7 +337,7 @@ selected_choice(void)
        size_t           yscroll = 0;
        int              dokey, doread, nready, timo;
        int              choices_reset = 1;
-       int              dochoices = 0;
+       int              dochoices = 1;
        int              dofilter = 1;
 
        cursor_position = query_length;
@@ -385,6 +387,10 @@ selected_choice(void)
                                            choices.length - length;
                                } else {
                                        choices_reset = 1;
+                                       /* XXX needed?
+                                       if (choices_count < choices_lines)
+                                               dofilter = 1;
+                                       */
                                }
                        } else {
                                nfds--; /* EOF */
@@ -484,6 +490,7 @@ selected_choice(void)
                        break;
                case LINE_DOWN:
                        if (selection < choices_count - 1) {
+                               dochoices = 1;
                                selection++;
                                if (selection - yscroll == choices_lines)
                                        yscroll++;
@@ -491,6 +498,7 @@ selected_choice(void)
                        break;
                case LINE_UP:
                        if (selection > 0) {
+                               dochoices = 1;
                                selection--;
                                if (yscroll > selection)
                                        yscroll--;
@@ -511,19 +519,24 @@ selected_choice(void)
                                yscroll = selection += choices_lines;
                        else
                                selection = choices_count - 1;
+                       dochoices = 1;
                        break;
                case PAGE_UP:
                        if (selection > choices_lines)
                                yscroll = selection -= choices_lines;
                        else
                                yscroll = selection = 0;
+                       dochoices = 1;
                        break;
                case END:
-                       if (choices_count > 0)
+                       if (choices_count > 0) {
                                selection = choices_count - 1;
+                               dochoices = 1;
+                       }
                        break;
                case HOME:
                        yscroll = selection = 0;
+                       dochoices = 1;
                        break;
                case PRINTABLE:
                        if (query_length == 0)
@@ -556,12 +569,14 @@ render:
                if (choices_reset)
                        choices_count = choices.length;
                choices_reset = 0;
-               if (dofilter) {
+               if (dofilter && !dochoices) {
                        dochoices = filter_choices(choices_offset,
                            &choices_count);
                        if (dochoices)
                                dofilter = 0;
                }
+               if (!dochoices && !dokey)
+                       continue;
 
                tty_putp(cursor_invisible, 0);
                tty_putp(carriage_return, 1);   /* move cursor to first column */
@@ -570,12 +585,13 @@ render:
                else
                        xscroll = 0;
                print_line(&query[xscroll], query_length - xscroll, 0, -1, -1);
+               if (selection >= choices_count)
+                       selection = yscroll = 0;
                if (dochoices) {
-                       if (selection >= choices_count)
-                               selection = yscroll = 0;
                        if (selection - yscroll >= choices_lines)
                                yscroll = selection - choices_lines + 1;
                        print_choices(yscroll, choices_count, selection);
+                       dochoices = 0;
                }
                tty_putp(carriage_return, 1);   /* move cursor to first column */
                for (i = j = 0; i < cursor_position; j++)
@@ -607,6 +623,7 @@ filter_choices(size_t offset, size_t *nchoices)
        size_t           i, match_length;
        size_t           n = 0;
        int              nready;
+       int              diff = 0;
 
        for (i = offset; i < *nchoices; i++) {
                c = &choices.v[i];
@@ -635,7 +652,17 @@ filter_choices(size_t offset, size_t *nchoices)
        qsort(choices.v, *nchoices, sizeof(struct choice), choice_cmp);
        *nchoices = offset + n;
 
-       return 1;
+       /* Determine if the order of visible choices did change. */
+       /* XXX does not respect yscroll */
+       for (i = 0; i < choices_lines; i++) {
+               if (i == choices.length)
+                       break;
+               if (choices_order[i] != choices.v[i].offset)
+                       diff = 1;
+               choices_order[i] = choices.v[i].offset;
+       }
+
+       return diff;
 }
 
 int
@@ -871,6 +898,12 @@ tty_size(void)
                tty_lines = 24;
 
        choices_lines = tty_lines - 1;  /* available lines, minus query line */
+
+       choices_order = reallocarray(choices_order, choices_lines,
+           sizeof(*choices_order));
+       if (choices_order == NULL)
+               err(1, NULL);
+       memset(choices_order, 0, choices_lines * sizeof(*choices_order));
 }
 
 void
```
... only `tests/opt-q.t` fails:
```sh
FAIL:	filter using the query option
CAUSE:	wrong output
--- stdout-want
+++ stdout-got
@@ -1 +1 @@
-a
+b
FAIL tests/opt-q.t (exit status: 1)
```
... but
```sh
echo "b\na" | LINES=5 ./pick -q a
```
on the command-line works as expected ...

dbotw:
@mptre this great PR should not be forgotten ...

dbotw:
@mptre please resurrect this PR!

PR #289, opened 2018-11-09
Fix build on illumos/Solaris

dbotw:
... there are still some warnings because of their system curses ... but it compiles and works!

`termios.h` is needed in `pick-test.c` for `TIOCSWINSZ`, but the call fails with `EINVAL`, which
seems to be a known bug: by ignoring the error most tests passed (only `key-alt-enter.t`
and `key-enter.t` timed out) ...

`/bin/sh` is `ksh93` there: no `local`-keyword, `diff` does not support the `-L` option ...

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -9,7 +9,7 @@
 #define __dead __dead2
 #endif /* __FreeBSD__ */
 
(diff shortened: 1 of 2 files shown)


mptre:
Thanks, committed in 2a5e047.

PR #290, opened 2019-02-20
Add support for building on Haiku/BeOS

dbotw:
`gcc 2.95` on Haiku/BeOS does not support `-Wextra`, `_BSD_SOURCE` is needed for `err`
and `_DEFAULT_SOURCE` to silence the compiler, that the previous macro is deprecated ...

`VDISCARD` is not in `POSIX.1` and (therefore?) not supported on Haiku/BeOS ...

`make check` does not work: `TIOCNOTTY` is "missing" ... no matter,  `test-suite.log` file
shows lots of  `pick-test: /dev/tt/p1: Operation not allowed` ...

Anyway, `pick` is running fine ...

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,7 @@
 AUTOMAKE_OPTIONS=foreign
 
-AM_CFLAGS=-Wall -Wextra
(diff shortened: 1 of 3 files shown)


mptre:
see #291

PR #291, opened 2019-02-20
Replace autoconf

mptre:
This has been on my list for a long while by now. Still some dependency issue with `make test` using GNU make, tried using [order-only-prerequisites](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html) without success.

Please give it a try on your platform.

/cc @calleerlandsson @mike-burns @dbotw @teoljungberg


dbotw:
Quoting https://www.gnu.org/software/make/manual/html_node/Suffix-Rules.html#Suffix-Rules:

```
Suffix rules cannot have any prerequisites of their own.
```

```diff
diff --git a/tests/Makefile b/tests/Makefile
index cfa3acb..4fd4978 100644
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -42,11 +42,11 @@ TESTS+=     opt-s.t
 TESTS+=        opt-unknown.t
 TESTS+=        opt-x.t
 
-all: ${TESTS:.t=.out}
+all: ${PROG} ${TESTS:.t=.out}
 
 .SUFFIXES: .t .out
 
-.t.out: ${PROG}
+.t.out:
        sh ${.CURDIR}/pick-test.sh -b ${.OBJDIR}/${PROG} $<
 
```

mptre:
On Thu, Feb 21, 2019 at 03:45:47AM -0800, Jens Guenther wrote:
> Quoting https://www.gnu.org/software/make/manual/html_node/Suffix-Rules.html#Suffix-Rules:
> 
> ```
> Suffix rules cannot have any prerequisites of their own.
> ```

Thanks for the pointer.

> 
> ```diff
> diff --git a/tests/Makefile b/tests/Makefile
> index cfa3acb..4fd4978 100644
> --- a/tests/Makefile
> +++ b/tests/Makefile
> @@ -42,11 +42,11 @@ TESTS+=     opt-s.t
>  TESTS+=        opt-unknown.t
>  TESTS+=        opt-x.t
>  
> -all: ${TESTS:.t=.out}
> +all: ${PROG} ${TESTS:.t=.out}

But there's no guarantee that PROG gets resolved before TESTS?

teoljungberg:
The deprecation of `-v` doesn't look like a deprecation, just a deletion. Semantically to me it's not the same to delete an option and to deprecate it, the latter displays a warning that it will be removed in the next major version bump.

teoljungberg:
I checked out the branch, ran the commands in CONTRIBUTING.md to build the project, with no avail:

```
pick autoremove % ./configure
pick autoremove % make
/Users/teo/src/pick/Makefile:1: /Users/teo/src/pick/Makefile.inc: No such file or directory
make: *** No rule to make target `/Users/teo/src/pick/Makefile.inc'.  Stop.
pick autoremove % make test
/Users/teo/src/pick/Makefile:1: /Users/teo/src/pick/Makefile.inc: No such file or directory
make: *** No rule to make target `/Users/teo/src/pick/Makefile.inc'.  Stop.
```

How do I generate `Makefile.inc`? That wasn't documented.

dbotw:
> The deprecation of -v doesn't look like a deprecation, just a deletion.

... and IMHO there is no good reason for this ...

dbotw:
> How do I generate Makefile.inc? That wasn't documented.

`./configure` should create the `Makefile.Inc`, but there is a problem with `sh` (bash 3.2) on Mac OS X:
the `config.log` file shows `./configure: line 19: @: unbound variable` ...

Changing line 19 in `./configure` to
```sh
$CC -Werror -o /dev/null -x c - "$@"
```
fixes this and `Makefile.Inc` should look like
```
CC=		cc
CFLAGS=		-Wall -Wextra -MD -MP
CPPFLAGS=	
DEBUG=		
INSTALL=	install
LDFLAGS=	-lcurses
MALLOC_OPTIONS=
```

dbotw:
```diff
diff --git a/tests/Makefile b/tests/Makefile
index cfa3acb..113a39a 100644
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -3,6 +3,7 @@ include ${.CURDIR}/../Makefile.inc
 PROG=  pick-test
 
 SRCS+= pick-test.c
+SRCS+= ../compat-reallocarray.c
 
 OBJS=  ${SRCS:.c=.o}
 DEPS=  ${SRCS:.c=.d}
@@ -42,7 +43,7 @@ TESTS+=       opt-s.t
 TESTS+=        opt-unknown.t
 TESTS+=        opt-x.t
 
-all: ${TESTS:.t=.out}
+all: ${PROG} ${TESTS:.t=.out}
 
 .SUFFIXES: .t .out
 
@@ -51,7 +52,7 @@ all: ${TESTS:.t=.out}
 
 
 ${PROG}: ${OBJS}
-       ${CC} ${DEBUG} -o ${PROG} ${OBJS} ${LDFLAGS}
+       ${CC} ${DEBUG} -o ${PROG} ${OBJS}
 
 clean:
        rm -f ${DEPS} ${OBJS} ${PROG}
```
... so `make test` works on Debian 9.7, Mac OS X and NetBSD 7.1.2 ...
(`LDFLAGS` removed: no need to link `pick-test` with `curses`)

teoljungberg:
I'd look into using https://github.com/koalaman/shellcheck or something similar to enforce `/bin/sh` compatibility.

mptre:
Thanks for the feedback, just pushed a few fixes:

* Keeping option `-v' for now, this is already a quite intrusive change.
  It might be deprecated in a correct manner as pointed out by Teo in
  the future. I think your package manager of choice already does a
  great job of keeping track of the installed version.

* Dependency issue with tests solved.

* Reworked Travis configuration, most notably re-adding support for
  sanitizers.

dbotw:
Tested and works ...

mptre:
> Tested and works ...

Thanks for testing, merged now.

configure
```diff
@@ -0,0 +1,227 @@
...
+HAVE_GNU_SOURCE=0
+HAVE_NCURSESW=0
+HAVE_NORETURN=0
+HAVE_PLEDGE=0
+HAVE_REALLOCARRAY=0
+HAVE_STRTONUM=0
+
+CC=$(makevar CC)
+CFLAGS=$(makevar CFLAGS)
+CFLAGS="${CFLAGS} ${DEBUG} -Wall -Wextra -MD -MP"
```
dbotw: Should `-MD` be replaced with `-MMD` to exclude system header?

mptre: Hm, why?

dbotw: Just to keep the dependencies files smaller than the source files:
```
$ wc -l compat-pledge.c
14 compat-pledge.c
$ wc -l compat-pledge.d
124 compat-pledge.d
```
with `-MMD`:
```sh
$ wc -l compat-pledge.d
3 compat-pledge.d
```
For a small project like `pick` dependencies in the Makefile would IMHO be enough ...

mptre:
> For a small project like `pick` dependencies in the Makefile would IMHO be enough ...

It might look a bit excessive. But, it's a matter of correctness. If a
system header changes, although highly unlikely, we want to rebuild
pick.

PR #294, opened 2019-03-08
Honor BINDIR during configure and install

mptre:
Ran into some problems while trying to package for OpenBSD. Turns out it's a better idea to only honor `PREFIX` during configure and during install always use dedicated variables for paths.

Does this work for everyone else?

/cc @dbotw @eavgerinos

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,8 +96,8 @@ distclean: clean
 .PHONY: distclean
 
 install: ${PROG}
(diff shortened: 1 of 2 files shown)


dbotw:
Works on NetBSD (and Debian) ...

mptre:
> Works on NetBSD (and Debian) ...

Thanks!

PR #299, opened 2019-12-12
Added multiple selections

nereusx:
* invoke pick with -m option to enable this ability
* press INSERT or CTRL+T to mark or unmark a line
* when you press [ENTER] now returns all selected items


pick.1
```diff
@@ -32,6 +32,8 @@ Both parts will be displayed but only the first part will be used when
 searching.
 .It Fl h
 Output a help message and exit.
+.It Fl m
```
mptre: Why is this feature opt-in to begin with?

pick.c
```diff
@@ -173,10 +184,22 @@ main(int argc, char *argv[])
 	choice = selected_choice();
 	tty_restore(1);
 	if (choice != NULL) {
-		printf("%s\n", choice->string);
-		if (output_description)
-			printf("%s\n", choice->description);
-	} else {
+		if ( opt_mark ) { /* if multiple choices, print out all selected strings */
```
mptre: If multiple choices are always enabled, you could drop the conditional and always iterate over all choices, printing the marked ones. In `case ENTER` just set `mark = 1` there as well then this logic should be applicable to all methods of chosing lines.

pick.c
```diff
@@ -851,7 +885,8 @@ print_choices(size_t offset, size_t selection)
 
 		if (i - offset < choices_lines)
 			print_line(choice->string, choice->length,
-			    i == selection, choice->match_start,
+					((i == selection) ? PL_STDOUT : 0) |\
```
mptre: Redundant parens, odd indent, redundant line continuation and redundant spaces below...

pick.1
nereusx: Without -m (multiple choices), the caller waits for one line string. in 'multiple choices'  mode the caller is prepared to receive a list of lines. So he had to specify which mode wants.

pick.c
nereusx: Yes, it cloud be and opt_mark = 0 case it could return immediately so no way to mess up with more marked lines.

pick.c
nereusx: I dont use spaces, only tabs, in a few cases spaces used to so more clear the one expression down of the other like this one. Try it with tab-width=4... Anyway, it is your code now rewrite it as you wish.

PR #306, opened 2020-04-29
Add RPM information to README.md

FreedomBen:


mptre:
When do you expect the RPM to appear in the official repo?

FreedomBen:
> When do you expect the RPM to appear in the official repo?

Really hard to predict.  Before it will be eligible for inclusion there are a number of things that need to be done, and several things depend on actions of other people outside of my control.  I would guess it will be "months" rather than "weeks" or "years."  

I'm planning to continue iterating toward it as I have time available.  Some of the things in the near future include signing the RPMs with my GPG keys, getting a COPR repo setup.  Before the package will even be considered for inclusion in the official repos I need to go through https://fedoraproject.org/wiki/Join_the_package_collection_maintainers and then https://fedoraproject.org/wiki/New_package_process_for_existing_contributors 

I love Fedora, but compared to other distros it is significantly more effort to package and ship software :-)

mptre:
Thanks for the reply. This will do for now then. Merged in be8676b.

Thanks for porting pick!

PR #312, opened 2020-12-27
Add COPR instructions for Fedora/RHEL/CentOS users

FreedomBen:
There is a COPR repo available now.  Many users prefer to avoid COPR
repos for various reasons so the old `wget` instructions are still
fully supported.

This is an important step toward getting included in the official Fedora
repos, so we're getting closer!  Can't put a timeline on inclusion yet
but I'm hoping the next major release will include pick directly.

There are _a lot_ of Fedora packaging standards/rules that I have to go
through and conform too, in addition to the standard process, but it's
coming along.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,18 @@ emerge pick
 ### Fedora/RHEL/CentOS
 
 The RPM builds are not yet in the official repositories, but you can install a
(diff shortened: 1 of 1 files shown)


mptre:
Thanks! Committed with some minor tweaks.

PR #316, opened 2022-11-11
fix build & configure for clang16

Schievel1:
Clang16 will not allow implicit function declarations by default which caused some trouble here. 

See also: https://bugs.gentoo.org/879771

Signed-off-by: Pascal Jäger <pascal.jaeger@leimstift.de>

### DIFF
--- a/configure
+++ b/configure
@@ -69,6 +69,7 @@ check_dead() {
 check_gnu_source() {
 	cat <<-EOF >"$TMP1"
 	#include <time.h>
+	int wcwidth(wchar_t wc);
 
 	int main(void) {
 		struct tm tm;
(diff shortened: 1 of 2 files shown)


mptre:
The whole purpose of the configure script is to sense presence of
certain functions. If the included header does not provide the wanted
function it will fail, which is expected and deliberatively handled.
It should therefore not define any prototypes at all.

This PR is a mess with all included patch leftovers.

Schievel1:
Yeah looks like I was focused a bit too much on getting rid of error messages last night so I didn’t pay much attention on what I’m doing.
That’s also why I made this a draft.

i will still leave this open for a while in case I see something different that needs fixing.

